PR #1, opened 2021-04-22
Toolchain

bobmcwhirter:
`make svd` will pull -rs, and populate the sources/svd/ dir.
Adjusted extract-all.sh to accomodate svd2rust being a peer/sibling to -metapac above -data.


Dirbaio:
Nice! Thanks

PR #4, opened 2021-04-28
Extract SYSCFG for L4 family.

bobmcwhirter:


data/registers/syscfg_l4.yaml
```diff
@@ -0,0 +1,306 @@
...
+      bit_size: 1
+    - name: ECCL
+      description: ECC Lock
+      bit_offset: 3
+      bit_size: 1
+    - name: SPF
+      description: SRAM2 parity error flag
+      bit_offset: 8
+      bit_size: 1
+fieldset/EXTICR1:
```
Dirbaio, line 103: `EXTICRx` should be an array, like in `syscfg_f4`. This way it's consistent :)

data/registers/syscfg_l4.yaml
```diff
@@ -0,0 +1,306 @@
...
+  description: SKR
+  fields:
+    - name: KEY
+      description: SRAM2 write protection key for software erase
+      bit_offset: 0
+      bit_size: 8
+fieldset/SWPR:
+  description: SWPR
+  fields:
+    - name: P0WP
```
Dirbaio, line 179: this should be an array.

data/registers/syscfg_l4.yaml
Dirbaio: Oops the regs already are an array, didn't see that. The fieldset should be `EXTICR` without the `1` then, and the unused other ones removed then. Like the F4 :)

PR #15, opened 2021-05-07
Helper script to extract a single peripheral from a single or all boa…

bobmcwhirter:
…rds, optionally with a peri-specific transform.yaml.


Dirbaio:
Great! I was missing it too, it got deleted somehow

PR #18, opened 2021-05-07
Add SPI v1.

bobmcwhirter:
All i2s status registers removed.

Renamed block to SPI from SPI1 for v2 (missed last time).


Dirbaio (approved):
:rocket:

PR #19, opened 2021-05-08
H7 SYSCFG

thalesfragoso:
I also changed `parse.py` to hand implement `removeprefix` and `removesuffix`


Dirbaio:
Looks nice! :) 

Good call about removeprefix/removesuffix, it's nice if the script works on older pythons :)

PR #28, opened 2021-05-21
Add peri-map for i2c v3.

bobmcwhirter:


Dirbaio:
Are you sure this is correct? according to [chibi](https://docs.google.com/spreadsheets/d/1-R-AjYrMLL2_623G-AFN2A9THMf8FFMpFD4Kq-owPmI/edit#gid=0) the F7 is i2c v2, not v3

bobmcwhirter:
I'll double check. Just the H7 regexp in parse and it grabbed F7 also. Please hold.

Dirbaio:
It's possible v2/v3's regs are really identical and chibi is separating them due to some other reason (or it's simply wrong). If that's the case then this PR is correct, but we should probably name it spi_v2.

bobmcwhirter:
Okay, they are identical, aside from documentation, where the F7 (v2) variant is better.  So I've removed v3, put an F7ish one for v2, changed the peri-map and regen'd, across 3 commits.

PR #41, opened 2021-06-08
Add missing clock selection for SPI6

lulf:
According to the reference manual, SPI6 is enabled in APB4, but the APB4Freq_Value peripheral list is not including SPI6.

### DIFF
--- a/data/chips/STM32H742AG.yaml
+++ b/data/chips/STM32H742AG.yaml
@@ -1430,6 +1430,7 @@ peripherals:
   SPI6:
     address: 0x58001400
     kind: SPI:spi2s2_v1_0
+    clock: APB4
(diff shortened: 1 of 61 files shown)


Dirbaio:
Is this hand-edited? it'll get overwritten the next time it's regenerated.

lulf:
> Is this hand-edited? it'll get overwritten the next time it's regenerated.

Yup. I can hack it into parse.py to set this if its h7 family and SPI6... Problem is the cubedb xml files are not correct.

Dirbaio:
Nice! Thanks for the PR :+1:

lulf:
@Dirbaio The fix now generates the correct clock for all H7 chips, which I somehow had missed in my initial commit anyway...

PR #48, opened 2021-06-20
Add EXTI and IPCC for STM32WB55

Tiwalun:
Add EXTI  for STM32WB55, IPCC for STM32WB55 and similar.

### DIFF
--- a/data/chips/STM32WB10CC.yaml
+++ b/data/chips/STM32WB10CC.yaml
@@ -98,6 +98,7 @@ cores:
     IPCC:
       address: 0x58000c00
       kind: IPCC:v1_0
+      block: ipcc_v1/IPCC
     IWDG:
       address: 0x40003000
       kind: IWDG:iwdg1_v2_0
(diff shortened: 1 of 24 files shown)


Dirbaio:
Could you arrayify all the possible registers? on IPCC the registers for core1 and core2 seem to be equal. On EXTI it seems everything can be arrayified too. 

In general we try to arrayify everything that can be, so that writing HALs is then much easier. When arrayifying it's OK to "uniformize" things a bit. For example EXTI RTSR1, RTSR2 have different bits implemented, but it's OK to make them just have all 32bits implemented, as if all 64 bits were implemented. This allows writing stuff like this in HALs: `exti.RTSR(n/32).write(|w| w.set_rt(n%32, true));` which is much nicer than big macros/matches.

Tiwalun:
Is it possible to create 2-dimensional arrays? 

For `EXTI`, I want to create an array out of `C1EMR1` and `C1EMR2`, but these could then be combined with `C2EMR1` and `C2EMR2` as well.

Dirbaio:
Not 2-dimensional, but you can create a "core" block that has one core's EMR, RTSR, FTSR, etc as arrays of len 2, then make an array of len 2 of that.

Dirbaio (approved):
Very nice! Thanks!

PR #49, opened 2021-06-22
F0 rcc, flash and syscfg

thalesfragoso:
Separated the RCC in two versions since they have slight different clock sources.


Dirbaio:
Thank you!

PR #53, opened 2021-06-28
Minor improvements

ctron:
There are two minor improvements:
* Ignore IntelliJ project files
* Check for the existence of some required commands before starting any processing
* Add a note in the readme on which tools are required and how to install `svd`


Dirbaio:
Very nice, thank you!

Issue #52, opened 2021-06-25
I2C missing in stm32l432kc


ctron:
According to the specs, the stm32l432kc should have an I2C interface. However it seems that this is not being generated.

And maybe I just didn't understand how the process works.

Dirbaio:
It does have the peripehrals, but with no block: https://github.com/embassy-rs/stm32-data/blob/main/data/chips/STM32L432KC.yaml#L160

It's missing an entry here https://github.com/embassy-rs/stm32-data/blob/main/parse.py#L308

According to [this](https://docs.google.com/spreadsheets/d/1-R-AjYrMLL2_623G-AFN2A9THMf8FFMpFD4Kq-owPmI/edit#gid=0) L4 should be i2c_v2. If you add it, compare the registers to make sure they're the same, just in case.

Dirbaio:
This has been fixed for a while already :)

PR #54, opened 2021-06-28
[#52]: Fix missing I2C

ctron:
I think I did the right thing. However, I have no real understanding of what I really did :grin: 

I also checked the `data/registery/i2c_v2.yaml` file with the PDF `dm00151940-stm32l41xxx42xxx43xxx44xxx45xxx46xxx-advanced-armbased-32bit-mcus-stmicroelectronics.pdf` … and it looks like the relative I2C registers and the absolute (for I2C1 and I2C3) match.

### DIFF
--- a/data/chips/STM32F030C6.yaml
+++ b/data/chips/STM32F030C6.yaml
@@ -81,6 +81,7 @@ cores:
       address: 0x40005400
       kind: I2C:i2c2_v1_1
       clock: APB1
+      block: i2c_v2/I2C
       pins:
       - pin: PB10
         signal: SCL
--- a/data/chips/STM32F030C8.yaml
+++ b/data/chips/STM32F030C8.yaml
@@ -81,6 +81,7 @@ cores:
       address: 0x40005400
       kind: I2C:i2c2_v1_1
       clock: APB1
+      block: i2c_v2/I2C
       pins:
       - pin: PB5
         signal: SMBA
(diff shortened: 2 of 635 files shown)


thalesfragoso:
Thank you, everything looks nice here, anything missing ?

Maybe one thing we should check is if the L5 and WL are using this i2c version too, since they got flagged.

ctron:
Ok, I will try to check and update this comment with my findings:

* [x] STM32L5
  * https://www.st.com/en/microcontrollers-microprocessors/stm32l5-series.html#documentation
  * Looks like the only manual is for STM32L552x and STM32L562xx
  * It has the same register mapping, but at the end, has a few additional registers:
    * `I2C_HWCFGR`, `I2C_VERR`, `I2C_IPIDR`, `I2C_SIDR`
    * Relative registers seem to map
    * Absolute addresses seem to match too
* [ ] STM32WL

ctron:
Looks good for me. Might be good to go.

Dirbaio:
>    I2C_HWCFGR, I2C_VERR, I2C_IPIDR, I2C_SIDR

They've added those to every peripehral in recent chips, I guess it's so that an OS can enumerate the present peripehrals at runtime. We can ignore those, they're not needed for operating the actual peripheral.

Dirbaio (approved):
Looks great, thank you!

For next PRs please do separate commits for manual and autogenerated changes, it makes review soooo much easier. Github UI explodes with 600 changed files...

PR #57, opened 2021-07-01
H7 clocksel

bobmcwhirter:


thalesfragoso:
Hmm, what happened to UART4 ?

bobmcwhirter:
It's weird on APB4 but a clocksel so not currently scraped from xml. 

Or am I missing something?

bobmcwhirter:
Lemme double check in the AM.

PR #69, opened 2021-07-28
Add F4 PWR

thalesfragoso:
This also adds the PWR peripheral (but not the block) for other chips that didn't have it on the XMLs.


Dirbaio:
:rocket:

PR #70, opened 2021-07-29
Include base addresses for flash and ram.

bobmcwhirter:
Remove 'null' entries for datasheets/reference-manuals.


Dirbaio:
Very interesting! I wonder if we should have separate fields for every "kind" of RAM?

bobmcwhirter:
Yah, I was starting to think about that.

Dirbaio:
Also maybe it's more natural to store the size in bytes, not in kbs. The ST XMLs store it in KBs but I find that quite strange, you can't represent smalle memory regions, like 512-byte for example

bobmcwhirter:
I'll keep poking on it and refining this PR.

Dirbaio:
Having size of each region is super nice, thanks! :) 

There's a few that don't make sense though. For example, on stm32f031c4 the total is less thank bank1. It seems the correct one is 16384.

```yaml
flash:
ram: 4
  bytes: 16384
  regions:
    BANK_1:
      base: 0x8000000
      bytes: 32768
```

bobmcwhirter:
I’ll investigate. The XMLs have a few configurations and xmltodict pruned
off those sub trees so it was sorta a challenge.

I’ll figure it out.

PR #71, opened 2021-07-30
Remove invalid bits from F4, L4 RCC

timokroeger:
Required for https://github.com/embassy-rs/embassy/pull/331

### DIFF
--- a/data/chips/STM32F410C8.yaml
+++ b/data/chips/STM32F410C8.yaml
@@ -254,7 +254,7 @@ cores:
     RCC:
       address: 0x40023800
       kind: RCC:STM32F410-rcc_v1_0
-      block: rcc_f4/RCC
+      block: rcc_f410/RCC
       pins:
       - pin: PC14
         signal: OSC32_IN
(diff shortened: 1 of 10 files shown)


Dirbaio:
I'm not sure this is right, the F410 does have RNGEN in AHB1ENR bit 31. The reason why they're both in the yaml is because it was probably made with `merge_regs.py`.

We probably have to split the rcc_f4 into many...

From the SVDs:

```
F410: AHB1ENR.RNGEN, bit 31
F411: no RNG
F412: AHB2ENR.RNGEN, bit 6
F413: AHB2ENR.RNGEN, bit 6
F427: AHB2ENR.RNGEN, bit 6
F429: AHB2ENR.RNGEN, bit 6
F446: no RNG
F469: AHB2ENR.RNGEN, bit 6

timokroeger:
The L4 bit indeed is known limitation from the SVD file:
https://github.com/stm32-rs/stm32-rs/blob/master/devices/stm32l4x1.yaml#L6-L10

Not sure why it’s not correctly patched… either I’m still doing something wrong or there is a bug in the stm32-rs scripts.

Dirbaio:
Oh I see! then the L4 change is correct, but the F4 still needs splitting

Dirbaio:
Awesome! :+1:

PR #72, opened 2021-07-30
STM32G0 support

bgamari:


Dirbaio:
I'm merging this even though you didn't mark it as ready for review, because otherwise it'll bitrot fast :)  Even if not complete, it's better to merge early and often, if it makes something new work.

Thanks for the PR!

bgamari:
Sure, no worries! I'm on holiday since the beginning of last week until the end of August so it would have been a while until I got around to carrying on with this work anyways. Moreover, the functionality that I implemented here appears to work.

PR #74, opened 2021-08-05
bxCAN Peripheral

timokroeger:
todo: check the bxcan1_v1_1_(F1|F7) variants


Dirbaio (approved):
Thank you!

timokroeger:
The current name "bxcan_v1" was just a first pick.
It might make fore sense to call it "can_bxcan" (and "can_bxcanf7" if it turns out to be a thing).
Then we can easily add the FDCAN peripheral later as "can_fdcan1", "can_fdcanX", ...
At the moment there is no chip that includes both bxCAN and FDCAN.

PR #76, opened 2021-08-13
add UART:sci2_v2_1 to parser

FrozenDroid:
Needed this to get the rest of the UARTs working on the STM32L476.

Is there a way to know which version a peripheral is from its "kind", other than guessing?


Dirbaio (approved):
Thank you!

> Is there a way to know which version a peripheral is from its "kind", other than guessing?

unfortunately no. The v1/v2/v3 is set by us based on which are register-wise compatible. The ST version number seems more granular, taking intou account interactions with clock/gpio/nvic stuff and maybe even silicon process / HDL version..? This makes it hard to map.

PR #81, opened 2021-08-18
Add wl55 radio spi

lulf:
This adds the peripheral block for SPI2s1 v3_5 and also adds a mapping for the SUBGHZSPI peripheral on the WL5x devices which is using the same register block.


Dirbaio (approved):
Thank you!

Funny that they gave it a special name to make it feel SpECiaL

PR #82, opened 2021-08-30
Fix ADC register layout on STM32G0

bgamari:
It appears that ST changed the register layout (particularly surrounding channel selection) quite drastically in the G0 series.

Note that the first commit captures changes that aren't due to this patch but rather were from simply rerunning `parse.py`.

Also added `PWR` support.


Dirbaio (approved):
Thank you! :rocket:

PR #83, opened 2021-08-31
Stm32wl55 pwr

lulf:
* Add PWR peripheral for STM32WL5*
* Handle SUBGHZ peripheral pins so that its properly recognized as a SPI peripheral by the macros
* Regenerate data


Dirbaio (approved):
:rocket:

Issue #10, opened 2021-05-01
Add stm32f1 GPIO AF maps


Dirbaio:
The F1 GPIO mapping works completely different, by "remapping" all the pins for a peripheral together. 

Invent some YAML format to represent that, and extract it from the GPIO XMLs.

DCNick3:
Copying over from #85:

> True... It seems they're all "compatible" in the sense that a given bit is either present or not present, but it never does X in one chip and Y in another chip. It's like the RCC's. In this case what we've done is "merge" all the yamls to obtain a single one with all the bits, that works for all the chips in the family. The merge_regs.py script does this.
> 
> #10 is about adding info on the YAMLs about pinouts. Something like "peripehral X with remap=0 uses pins XXX, with remap=1 uses pins YYY". Since it works differently than all the other stm32 families it'll need a new YAML format.
> 
> This info is then used in embassy-stm32 to autogenerate the trait implementations to enforce correct pins when creating an UART etc.

DCNick3:
This seems to be possible, but _may_ require API changes. Because in F1 you can't change pin functions independently, the pin trait bounds like `SclPin<T: Instance>` would not work as-is, they would rather need to be defined on tuples of pins, not on individual ones.

PR #85, opened 2021-09-22
Add initial register mapping for STM32 F1 AFIO and FLASH

mryndzionek:
Added just the register needed for STM32 F1 support in Embassy. Not sure if `afio_f1` should be `afio_v1` (AFIO is needed for EXTI control for `gpio_v1`, so renaming probably makes sense).

### DIFF
--- a/data/registers/afio_f1.yaml
+++ b/data/registers/afio_f1.yaml
@@ -1,0 +1,141 @@
+---
+block/AFIO:
+  description: Alternate function I/O and debug configuration
+  items:
+    - name: EVCR
+      description: event control register
+      byte_offset: 0
+      reset_value: 0
+      fieldset: EVCR
+    - name: MAPR
+      description: AF remap and debug I/O configuration register
+      byte_offset: 4
+      reset_value: 0
+      fieldset: MAPR
+    - name: EXTICR
+      description: external interrupt configuration register
+      array:
+        len: 4
+        stride: 4
+      byte_offset: 8
+      reset_value: 0
+      fieldset: EXTICR
+    - name: MAPR2
+      description: AF remap and debug I/O configuration register2
+      byte_offset: 28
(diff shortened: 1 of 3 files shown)


Dirbaio:
Thanks for the PR!

The only issue I see is  FLASH register block is not complete though. We shouldn't have incomplete blocks, as it can create confusion. 

You can extract them automatically from SVD files following the instructions [here](https://github.com/embassy-rs/stm32-data#extracting-new-register-blocks).

> Not sure if afio_f1 should be afio_v1 (AFIO is needed for EXTI control for gpio_v1, so renaming probably makes sense).

AFIO only exists on F1 anyway, so either is fine, I guess

mryndzionek:
Okay, I updated the FLASH registers. I also noticed that AFIO is not mutually consistent (ties probably with #10).

Dirbaio:
Thank you!

> I also noticed that AFIO is not mutually consistent (ties probably with #10).

True... It seems they're all "compatible" in the sense that a given bit is either present or not present, but it never does X in one chip and Y in another chip. It's like the RCC's. In this case what we've done is "merge" all the yamls to obtain a single one with all the bits, that works for all the chips in the family. The `merge_regs.py` script does this.

#10 is about adding info on the YAMLs about pinouts. Something like "peripehral X with remap=0 uses pins XXX, with remap=1 uses pins YYY". Since it works differently than all the other stm32 families it'll need a new YAML format.

This info is then used in embassy-stm32 to autogenerate the trait implementations to enforce correct pins when creating an UART etc.

mryndzionek:
Okay, so I think I need to re-generate AFIO yaml. The one here was done manually. Do you want me to do this while I'm at it?

Dirbaio:
Sure, why not! :)

mryndzionek:
Okay, PR #87

PR #88, opened 2021-09-24
Add L4 flash register

FrozenDroid:
This adds the flash register for L4 devices, needed for a PR I'm about to make in embassy for additional L4 clock sources.


Dirbaio (approved):
Thank you!

PR #89, opened 2021-09-25
F4 CRC peripheral

theunkn0wn1:
> v1: F1, F2, F4, L1
v2, adds INIT reg: F0
v3, adds POL reg: F3, F7, G0, G4, H7, L0, L4, L5, WB, WL

This PR Does two things:
1. fix CRC being treated as a software library; the families implement this in hardware.
2. Adds three versions of the CRC peripheral, as per discussion  in matrix.


Dirbaio (approved):
Looks awesome, thank you!

PR #90, opened 2021-09-25
Added AFIO block definition to STM32F103C8 chip

mryndzionek:


lulf:
Did you edit the parse.py as well or hand edit the YAML? We want to avoid hand-editing the YAML, because it gets overwritten if someone changes parse.py.

mryndzionek:
> Did you edit the parse.py as well or hand edit the YAML? We want to avoid hand-editing the YAML, because it gets overwritten if someone changes parse.py.

Adding AFIO was done in #87. YAML in this PR was hand-edited. Should I re-generate it somehow?

lulf:
> > Did you edit the parse.py as well or hand edit the YAML? We want to avoid hand-editing the YAML, because it gets overwritten if someone changes parse.py.
> 
> Adding AFIO was done in #87. YAML in this PR was hand-edited. Should I re-generate it somehow?

Yup, you need to run `python3 parse.py` to regenerate the chip definitions. This requires running ./d download-all  to ensure you have all the sources from which the YAML is generated.

mryndzionek:
I did `download-all`. If I run `python3 parse.py` there is a huge diff generated - mostly `DMAMUX2` changed 'position' in all the YAML files. `AFIO` is however not added to `data/chips/STM32F103C8.yaml`. Am I missing something more?

lulf:
The DMAMUX changes seems to be some python list/dict ordering issue. I had a look at the cubedb XML, and I can't find the AFIO peripheral there, so I think it must be handled similarily to PWR, DBGMCU etc. You can find the code for these peripherals here https://github.com/embassy-rs/stm32-data/blob/main/parse.py#L973-L994

So you need to do the same for AFIO. I'll see if I can resolve the ordering issue so you can ignore that for now.

mryndzionek:
I think I managed to do what is needed the way you described. Let me know if this is okay.

PR #92, opened 2021-10-19
Initial support for STM32F767ZI.

matoushybl:
This PR adds support for the STM32F767ZI, it only adds basic peripherals.

I am looking forward to your feedback on improving it and adding support for more peripherals and devices in the F7 family.


Dirbaio:
`chips/*.yaml` are autogenerated by `parse.py`, please do not edit them manually. 

You'll want to add the new peripehrals [here](https://github.com/embassy-rs/stm32-data/blob/869cd60e9d2cc17543c81394dc04ba6b1eb9dac2/parse.py#L311) and then run `parse.py` to update all the yamls.

(please make separate commits with manual vs autogenerated changes, it helps a lot with review!)

Dirbaio:
> Are you in favor of adding a bit more information to the README about it as well?

Yes, of course! The current docs are lacking, indeed.

matoushybl:
Also how does the generator handle the RAM region patching? I didn't find it in the `parse.py` file, but maybe I haven't looked hard enough.

Dirbaio:
The memory map parsing is wrong for many families, not just F7. The XMLs have a "ram size" and parse.py assumes it's the size of SRAM, but it is usually the total (SRAM1+SRAM2+DTCM+etc). 

It needs some general refactor to make it correct, probably either finding more detailed data somewhere, or feeding it in ourselves... I would ignore it for this PR for now. You can workaround it by not enabling the `embassy-stm32` Cargo feature `memory-x`, and supplying your own `memory.x`.

PR #99, opened 2021-11-08
More support for STM32U5

bobmcwhirter:


bobmcwhirter:
Yes, I do realize RCC enums are still a bit wack, but I'll continue to iterate through them.

Dirbaio:
Thank you! :rocket:

PR #112, opened 2021-12-23
Assign correct register blocks for F3 Timers

VasanthakumarV:
Previously all `TIMx` registers in F3 chips were assigned to `timer_v1/TIM_GP16`, this PR maps `TIMx` registers to their correct register blocks.


Dirbaio (approved):
Neat, thank you! :)

PR #115, opened 2021-12-29
stm32f1xx adc support

sjoerdsimons:
Add the various bit need to support ADC on a stm32f1xx (tested on a blue pill).


Dirbaio:
Thank you!

PR #117, opened 2022-01-14
Unify SPI LSBFirst enums.

matoushybl:
This leaves out some of the cfg attributes needed for SPI setup, alongside the ugly conditional includes.


Dirbaio (approved):
Neat! :rocket:

Issue #120, opened 2022-01-25
Add LPUART support (for L4)


mro95:
Hi, I would like to begin adding support for LPUART in the HAL. 
What is the workflow to add a new peripheral to the meta-pac?

One of my issues was that the `./d extract-all` script doesn't match `LPUART` but does match `LPUART1`

Dirbaio:
> One of my issues was that the ./d extract-all script doesn't match LPUART but does match LPUART1

This is expected, it matches by peripheral name, which is LPUART1. It's not an issue, just change `block/LPUART1:` to `block/LPUART:` in the yaml afterwards.

- Run `./d extract-all LPUART1`. This'll output a bunch of yamls in `tmp/LPUART1`
- Diff them between themselves, to identify differences. The differences can either be:
   - 1: Legitimate differences between families, because there are different LPUART versions. For example, added registers/fields.
   - 2: SVD inconsistencies, like different register names for the same register
   - 3: SVD mistakes (yes, there are some)
   - 4: Missing stuff in SVDs, usually enums or doc descriptions.
- Identify how many actually-different (incompatible) versions of the peripheral exist, as we must *not* merge them. Name them v1, v2.. (if possible, by order of chip release date, see [here](https://docs.google.com/spreadsheets/d/1-R-AjYrMLL2_623G-AFN2A9THMf8FFMpFD4Kq-owPmI/edit#gid=1972450814).
- For each version, pick the "best" yaml (the one that has less enums/docs missing), place them in `data/registers/lpuart_vX.yaml`
- Add entries to [`perimap`](https://github.com/embassy-rs/stm32-data/blob/main/stm32data/__main__.py#L84)
- Regen, check `data/chips/*.yaml` has the right `block: lpuart_vX/LPUART` fields

(do not use merge-regs.py, that's only for RCC)

mro95:
Thanks for the help! As you can see, I made a PR for this.
Because those things are not documented, is it possible to add your last comment (maybe with some changed) in the readme of wiki?

Dirbaio:
Done! :)

PR #121, opened 2022-01-26
Add LPUART peripherals

mro95:
As I mentioned in issue #120, I would like to begin adding support for LPUART.
I followed the instructions of @Dirbaio

Let me know what you think

### DIFF
--- a/data/chips/STM32G031C4.yaml
+++ b/data/chips/STM32G031C4.yaml
@@ -464,6 +464,7 @@ cores:
           reset:
             register: APBRSTR1
             field: LPUART1RST
+      block: lpuart_v2/LPUART
       pins:
       - pin: PA2
         signal: TX
--- a/data/chips/STM32G031C6.yaml
+++ b/data/chips/STM32G031C6.yaml
@@ -464,6 +464,7 @@ cores:
           reset:
             register: APBRSTR1
             field: LPUART1RST
+      block: lpuart_v2/LPUART
       pins:
       - pin: PA2
         signal: TX
--- a/data/chips/STM32G031C8.yaml
+++ b/data/chips/STM32G031C8.yaml
@@ -464,6 +464,7 @@ cores:
           reset:
             register: APBRSTR1
             field: LPUART1RST
+      block: lpuart_v2/LPUART
       pins:
       - pin: PA2
         signal: TX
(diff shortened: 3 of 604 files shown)


mro95:
I noticed that there is no DMA block, did i forget something to do?

Dirbaio:
Thanks for the PR!

There's actually 4 different versions of LPUART according to the SVDs (I've done the diffing, added them to the [table](https://docs.google.com/spreadsheets/d/1-R-AjYrMLL2_623G-AFN2A9THMf8FFMpFD4Kq-owPmI/edit#gid=1972450814))

- v1: L0, L4, L4+
- v2: G0, G4, L5, WL -- all regs from v1, plus PRESC
- v3: H7, WB -- all regs from v1, plus PRESC, GTPR, RTOR
- v4: U5 -- all regs from v1, plus PRESC, AUTOCR (yes, removes GTPR, RTOR :shrug: )

You've mapped all chips to version v1, which s not right. You should either add all versions, or If you personally only care about L4 support, add v1 only to L0, L4, L4+ ) and I'll add the others. Up to you :) 

(The reason I'd rather not have "wrong" versions is that later on it's much easier to catch "there's no LPUART for U5" than "there's LPUART for U5 but it's missing registers")

Dirbaio:
> I noticed that there is no DMA block, did i forget something to do?

I think it's because ST calls the DMA request `LPUART` while the peripheral is `LPUART1`, so the script doesn't match them: https://github.com/embassy-rs/stm32-data-sources/blob/main/cubedb/mcu/IP/DMA-STM32L4A6_dma_v1_1_Modes.xml#L720

An `if` in the right place in the script should solve it...

(ST loves to name the peripheral FOO instead of FOO1 when there's just one, and worst of all is they make the renames inconsistenly! :angry: . I hope to someday make the scripts normalize FOO to FOO1 so that naming is consistent between chips that have 1 foo and many foos, but that's a battle for another day...)

Dirbaio:
Thanks for working on this!

Turns out GTPR, RTOR existing is a mistake in the SVDs, so v2 and v3 are really the same... I've merged them.

I've fixed up v1 and v2 so the diff between them is as small as possible, and removed useless "0=disabled 1=enabled"  enums.

Dirbaio:
For some reason I can't push to this branch, I've manually merged this + my fixups to main.

PR #122, opened 2022-02-03
Fix missing USB OTG pins

chemicstry:
This fixes missing pin numbers in peripherals that have underscores in their names. Specifically `USB_OTG_FS` and `USB_OTG_HS`.

Also fixed paths in windows. The glob function returns paths with backward slashes.

I did not regenerate files, because I also get weird output unrelated to these changes. I gave editing permission for maintainers to do so.

### DIFF
--- a/stm32data/__main__.py
+++ b/stm32data/__main__.py
@@ -662,9 +662,14 @@ def parse_chips():
                         signal = 'SUBGHZSPI_' + signal[16:-3]
                     # TODO: What are those signals (well, GPIO is clear) Which peripheral do they belong to?
                     if signal not in {'GPIO', 'CEC', 'AUDIOCLK', 'VDDTCXO'} and 'EXTI' not in signal:
-                        periph, signal = signal.split('_', maxsplit=1)
-                        pins = periph_pins.setdefault(periph, [])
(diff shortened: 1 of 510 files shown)


stm32data/__main__.py
```diff
@@ -438,7 +438,11 @@ def cleanup_pin_name(pin_name):
 
 
 def parse_signal_name(signal_name):
-    parts = signal_name.split('_', 1)
+    if signal_name.startswith('USB_OTG_FS') or signal_name.startswith('USB_OTG_HS'):
```
chemicstry, line 441: I couldn't do a clever solution like in `parse_chips()`, because peripheral names are not available at this point. Would it make sense to change `parse_chips` function into the same `if` statements for faster processing? TBH, I did not notice any slowdown due to array lookup.

chemicstry:
I think this is *finally* ready.

chemicstry:
Fixed json encoding on windows so regenerated all files

stm32data/__main__.py
Dirbaio: If it's not slower then it's fine :)

Dirbaio (approved):
This is great, thank you for working on it! Thanks for the windows fixes as well :)

PR #123, opened 2022-02-28
Use "CAN" for chips with a single instance

timokroeger:
Required for the embassy trait implementations to be correct:
https://github.com/embassy-rs/embassy/blob/35636953b2048394733eaff95fad17970788d08a/embassy-stm32/src/can/bxcan.rs#L95

### DIFF
--- a/data/chips/STM32F722IC.json
+++ b/data/chips/STM32F722IC.json
@@ -614,7 +614,7 @@
                     }
                 },
                 {
-                    "name": "CAN1",
+                    "name": "CAN",
                     "address": 1073767424,
                     "registers": {
                         "kind": "can",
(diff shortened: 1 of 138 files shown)


Dirbaio:
Do not manually edit the chip JSONs, they're autogenerated. See README.

Please modify the script to do what you want, then do a first commit with just the script changes, and a second commit with the result of rerunning it.

Dirbaio:
Also, It feels a bit fragile to me to have embassy rely on "CAN" vs "CAN1" to deduce the capabilities. No other diver does that at the moment. 

Maybe it would be better to add cfg's on the embassy side so each chip gets the right capabilities? USB_OTG does this.

timokroeger:
Closing for the reasons described by Dirbaio. I will have a look at the cfg solution in embassy-stm32.

PR #124, opened 2022-03-02
STM32G0 updates

msamsonoff:
Additions and fixes for the G0 family.

NOTE: The addition of enums for the RCC breaks embassy-stm32/src/rcc/rcc_g0.rs. I am preparing a fix for that.


Dirbaio:
This is awesome, thank you! :)

PR #127, opened 2022-03-08
Don't rename DMA channels if they don't start at zero

GrantM11235:


Dirbaio (approved):
Awesome, thank you! :)

PR #129, opened 2022-03-15
Timers use the clock speed apb1_tim and apb2_tim

nviennot:
Timers need to use this clock definition: https://github.com/embassy-rs/embassy/blob/cb1be3983a47029e0e6a1c07082dd6ed18eda2a7/embassy-stm32/src/rcc/mod.rs#L29

I'm not entirely sure this is okay for all STM32s


Dirbaio (approved):
Thank you!

PR #130, opened 2022-03-16
Unify SDMMC v1 and v2 register names

chemicstry:
SDMMC v2 is an extension of SDMMC v1 peripheral so most of the registers are the same. This PR fixes minor inconsistencies between register names.

This also assigns SDIO peripheral to SDMMC v1 as they are identical (?).

Keeping this as draft now, until SDMMC v1 driver is ready for embassy.


Dirbaio:
Thank you!

PR #132, opened 2022-03-17
F2: Add SYSCFG/FLASH/PWR, fix DBGMCU SVD typos

Gekkio:
I've got a project using STM32F215 and am working on contributing possible F2 support to embassy :smile:
Here's the first batch of improvements, which hopefully will be enough to add basic support for this family in embassy.

After these changes, at least the following F2 peripherals are still missing proper register definitions:

* DCMI
* CRYP
* IWDG
* HASH
* RTC
* SPI1
* SPI2
* SPI3
* SDIO
* ETH
* FSMC

I'll probably add at least SPI and FSMC in a later pull request since I'm going to need those :smile:

### DIFF
--- a/data/chips/STM32F205RB.json
+++ b/data/chips/STM32F205RB.json
@@ -942,6 +942,11 @@
                 {
                     "name": "FLASH",
                     "address": 1073888256,
+                    "registers": {
+                        "kind": "flash",
+                        "version": "f2",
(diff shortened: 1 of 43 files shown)


Dirbaio:
Oooh, F2, exciting. Thanks for the PR, looking forward to the next ones!

PR #133, opened 2022-03-19
FSMC register block

nviennot:
The current v1 yaml has a few extra registers defined that don't belong to some of the chips out there (the ones at byte_offset 320 and above), but the rest of registers are identical.


Dirbaio:
Thank you! :rocket:

PR #134, opened 2022-03-20
Add USB_OTG_FS registers

nviennot:


data/registers/otgfs_v1.yaml
```diff
@@ -1,140 +1,297 @@
...
+      description: Host periodic transmit FIFO size register
       byte_offset: 256
-      fieldset: OTG_FS_HPTXFSIZ
-    - name: OTG_FS_DIEPTXF1
-      description: OTG_FS device IN endpoint transmit FIFO size register (OTG_FS_DIEPTXF1)
+      fieldset: HPTXFSIZ
+    - name: DIEPTXF
+      description: Device IN endpoint transmit FIFO size register
+      array:
+        len: 3
```
Dirbaio, line 90: do all chips have max 3 endpoints? iirc some have more, we should set this to the max, the HAL should know the actual max endpoint count.

nviennot: From what I see in the extracted yamls (./d extract-all OTG_FS_{GLOBAL,HOST,DEVICE,PWRCLK}):
* The f101 f102 f103 f107 f215 f217 f412 f413 l4x6 families: all have only 3 endpoints
* The f469 family has 5 endpoints. Its datasheet shows High-Speed support, not just Full-Speed. From a quick glance I don't see extra registers in the yaml.
* The f730 f745 f750 f765 f7x2 f7x3 f7x6 f7x7 f7x9 families: 5 endpoints, but they also have a bunch of additional registers like GI2CCTL GLPMCFG GPWRDN GADPCTL. So that should be a separate yaml definition
Overall, I suspect that the fanciers cores with more endpoints have extra features, and deserve their own v2 yaml.
At this point, it's probably not harmful to keep the endpoint at 3. The fancier cores will be limited, yes. But to give them justice we'll have to do better than just bumping the number of endpoints.

Dirbaio:
LGTM, Thank you!

PR #135, opened 2022-03-24
WIP: add usb registers to l4r5

ant32:
I got usb working on STM32L4R5ZITx

I followed this example https://github.com/embassy-rs/embassy/blob/5c68f0bae7c4091ad34fb2a671e08a614d9beb9a/examples/stm32f4/src/bin/usb_uart.rs but made some changes
```rust
// USB requires at least 48 MHz clock
fn config() -> Config {
    let mut config = Config::default();
    config.rcc.mux = ClockSrc::MSI(MSIRange::Range11);
    config
}

#[embassy::main(config = "config()")]
async fn main(_spawner: Spawner, p: Peripherals) {
    // Enable PWR peripheral
    unsafe { RCC.apb1enr1().modify(|w| w.set_pwren(true)) };

    // Enable VddUSB
    let pwr_cr2 = 0x40007004 as *mut i32;
    unsafe { *pwr_cr2 |= 1 << 10 }; // USV is on bit 10;
```

### DIFF
--- a/data/chips/STM32F722IC.json
+++ b/data/chips/STM32F722IC.json
@@ -5472,6 +5472,11 @@
                 {
                     "name": "USB_OTG_FS",
                     "address": 1342177280,
+                    "registers": {
+                        "kind": "otgfs",
+                        "version": "v3",
+                        "block": "OTG_FS"
+                    },
                     "rcc": {
                         "clock": "AHB2",
                         "enable": {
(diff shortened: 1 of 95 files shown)


Dirbaio:
The chip JSON files are autogenerated.

Could you edit the Python code to generate what you need instead? then make one commit with the changes to the python code, and another with the results of running it?

Dirbaio:
Thank you! :)

ant32:
Sorry, I was going to spend a bit more time on this and had tried marking it as Work In Progress. My example code isn't compiling yet with this. I'll try figure it out next week and create another pull request

PR #136, opened 2022-03-29
Use OTG v1 for v3

ant32:
I can not find any meaningful differences between v1 and v3. The initial pull request was missing the HOST, DEVICE, PWRCLK blocks and wasn't cleaned up for the transform-USB_OTG.yaml transformations. After carefully making those changes I was mostly left with fieldset names not matching. I only looked at the l4r9 yaml today while merging. I noticed that the f7 had some additional registers but didn't take any time to study them. The u5xx has some much longer descriptions that looked interesting but also didn't have descriptions at all for many.

From what I could find cubedb has the following series marked as otg v3
- f7x2
- f7x3
- f730
- l4r9 (stm32l4 plus series)
- u5xx

here is one register I found that was missing on the l4+ reference manual
```yaml
    - name: EPMISM
      description: Endpoint mismatch interrupt mask
      bit_offset: 17
      bit_size: 1
```


Dirbaio:
Thank you!

PR #137, opened 2022-03-29
Add missing timer ITR3 field

chemicstry:


Dirbaio:
Thank you!

(these seem like they could be an array though)

PR #138, opened 2022-04-09
add stm32l4 power registers

ant32:
None of the l4 yaml had enums so I added them from wl5 and double checked all of them with the reference manual. I can squash the first 3 commits if needed.


Dirbaio:
Thank you!

PR #139, opened 2022-04-26
Stm32wl flash

lulf:


Dirbaio:
could you name it `flash_wl`? it's also used for WLEx which is not WL55x

Dirbaio:
Thank you! :rocket:

PR #140, opened 2022-04-26
Generate support for ethernet v1a and v1b

davidlenfesty:
I pulled out what wasn't used in these two according to `stm32-rs`, haven't verified against the datasheets, but drivers relying on them aren't using any of the registers with differences and so should work for all versions.


Dirbaio:
Looks great, Thank you! :rocket:

PR #143, opened 2022-05-02
Clean up rcc for f1/f1cl and separate f100 rcc

GrantM11235:


Dirbaio:
This is great work, thanks for dealing with this! 

(why is F1 RCC so horrible??)

PR #145, opened 2022-05-10
Add LCD peripheral

jensenn:
This creates register access to the LCD controller peripheral. This was tested on a STM32 L1 Discovery board. https://www.st.com/en/evaluation-tools/32l152cdiscovery.html

I'm not sure a single bit addition in a register is enough to warrant a v2 register set, but I ended up creating it.


Dirbaio:
Thank you!

PR #146, opened 2022-05-18
registers/can_bxcan: change filter register from bitfield to u32

davidlenfesty:
This is currently bugging me :)


davidlenfesty:
Er, not sure what's correct for the FB length/stride. Pattern matching is failing me right now, debugging other CAN issues so my brain's a bit fried.

The actual layout is 14 or 28 banks of 2 32 bit registers, depending on the model. Not sure how to handle product line splits, the easiest way is to just say it's always 28 and assume people are looking at the correct reference manual.

Dirbaio:
> the easiest way is to just say it's always 28 and assume people are looking at the correct reference manual.

Yes, for these cases (register layout is the same, but there's a different amount of $THING) we set it to the max amount and leave it up to the HAL to know the amount for the chip.

> Er, not sure what's correct for the FB length/stride. Pattern matching is failing me right now, debugging other CAN issues so my brain's a bit fried.

I think the old len+stride was correct. There's chips with up to 28 FBs (stm32f4), and stride is in bytes: 32bit x 2 = 8 bytes.

About removing the bitfields: you can get/set the raw `u32` from a `Fr1/Fr2` by doing `.0`, so maybe it's best to keep them? This way it matches the RM, and user code can still access it both ways (bitwise, or the entire raw `u32`)

PR #147, opened 2022-05-31
Add RTC registers for all chips

chemicstry:
The v1 (F1) and v3 (G0, G4, L41x, L42x, WL) versions are straightforward.

The v2 version, however, is not consistent at all. Even for the same peripheral version (i.e. `rtc2_v2_4_Cube`) the registers have differences for each series (F0, F3, L4). So I split them into different files for each series. Is there a better way to do it?


chemicstry:
So I think this is finally ready for review...

I manually compared all versions of the peripheral from SVDs and made sure that all the files are mostly the same, except for the few register differences between series. No doubt there will be errors (i.e. differences between chips inside the series), but this is the best I can do without double checking each reference manual. I hope they can be ironed out when/if HAL driver is written.

Dirbaio:
Huge thanks for the huge work @chemicstry! it's crazy that there are so many v2 variants, what the hell? Looks good overall, just a few nits:

- RTC on L41x/L42x is different than on the other L4s (it's a v2 variant I think, not v3)
- TAMPx bits on TAMPIE could be arrayified.

chemicstry:
> * RTC on L41x/L42x is different than on the other L4s (it's a v2 variant I think, not v3)

Yes, but cube xmls specify a correct version for those and they are handled by `.*:RTC:rtc3_v2_0` selector, while other L4's are handled by `STM32L4.*:RTC:rtc2_.*`.

> * TAMPx bits on TAMPCR could be arrayified.

Done! Also arrayified TAMPxIE, TAMPxNOERASE, TAMPxMF, but replaced the number an `X` as `TAMPxIE` field would clash with existing `TAMPIE` (global interrupt enable). Is this ok?

Dirbaio:
> Yes, but cube xmls specify a correct version for those and they are handled by .*:RTC:rtc3_v2_0 selector, while other L4's are handled by STM32L4.*:RTC:rtc2_.*.

Output JSONs for `L4[12]` end up with `v3` though: https://github.com/chemicstry/stm32-data/blob/rtc_regs/data/chips/STM32L412C8.json#L1556

Dirbaio:
Rebased

Dirbaio:
Thanks again for the hard work! :D

PR #148, opened 2022-06-03
Add F1 BKP peripheral

chemicstry:
This is an addition to #147, but made it a separate PR to keep it less clutered.

F1 chips have a separate BKP peripheral, which contains the backup data registers and also some extra RTC stuff.


chemicstry:
Oh, and it wouldn't be ST if data registers were continuous. So I named them `DR1` and `DR2`, unless arrays somehow support non-contiguous layout?

chemicstry:
Done! Typed all 42 of them 😲

Dirbaio (approved):
Thank you!! :rocket:

PR #149, opened 2022-06-21
Fix TIM timer types across all lines

smbolton:
This PR fixes TIM timer types (TIM_ADV, TIM_GP32, TIM_BASIC, TIM_GP16) across all lines. The resulting assignments match both [AN4013 Application note - STM32 cross-series timer overview](https://www.st.com/resource/en/application_note/an4013-stm32-crossseries-timer-overview-stmicroelectronics.pdf) and the current set of SVDs. For the following part numbers, AN4013 and the respective SVD didn't clearly agree, in which case the datasheets were consulted for confirmation:

    STM32L4R9
    STM32L552
    STM32L562
    STM32G0* except for STM32G070

(In each of these TIM3 and/or TIM4 inherits from TIM2 and the GP32/GP16 distinction isn't clear in the SVD.)


Dirbaio:
This is really great, thank you!

PR #153, opened 2022-08-03
DAC update and cleanup

chemicstry:
- Fixed DAC register mapping for F1
- Arrayified DAC registers where possible (channels 1/2)
- Removed "enable" enums
- Removed all non-functional differences between v1 and v2 (check diff between them)

Technically v2 DAC is just an extension of v1, so I'm not sure if it's worth keeping 2 separate definitions? For other peripherals that is handled in the HAL. The differences are:
- v1 has output buffer disable bit, where v2 doesn't
- v2 has additional calibration, mode, sample hold registers

### DIFF
--- a/data/chips/STM32F100C4.json
+++ b/data/chips/STM32F100C4.json
@@ -421,6 +421,11 @@
                 {
                     "name": "DAC",
                     "address": 1073771520,
+                    "registers": {
+                        "kind": "dac",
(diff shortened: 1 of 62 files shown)


Dirbaio:
LGTM, thank you! :)

re v1/v2: If the regs are different, it's better to keep the versions separate.

PR #154, opened 2022-08-18
USARTv2 split ixr into icr and isr

FrozenDroid:
IXR was incorrect due to a few bits not being writable (or readable, even) in ICR


Dirbaio (approved):
Thank you! nice catch :)

PR #155, opened 2022-08-19
Fix LPUART interrupts for the L0 family

FrozenDroid:
Slight typo in the source files which caused the interrupts to not be matched correctly.


Dirbaio (approved):
Thank you for your work! :rocket:

PR #157, opened 2022-09-09
Special case handling for STM32F0 ADC peripheral

### LINKED ISSUE #156: Wrong base address for ADC on STM32F0
The JSON files have the wrong base address for the ADC peripheral for the STM32F0 family. This includes both the RM0091 parts and the RM0360 parts.

    "peripherals": [
        {
            "name": "ADC",
            "address": 1073817352,
            "registers": {
                "kind": "adc",
                "version": "v1",

The address value should be `1073816576` (`0x40012400`.)

One possible fix is more special-casing code:

    diff --git a/stm32data/__main__.py b/stm32data/__main__.py
    index a328a94da..d71eef4f2 100755
    --- a/stm32data/__main__.py
    +++ b/stm32data/__main__.py
    @@ -769,7 +769,10 @@ def parse_chips():

                 peris = []
                 for pname, pkind in peri_kinds.items():
    -                addr = get_peri_addr(defines, pname)
    +                if chip_name.startswith('STM32F0') and pname == 'ADC':
    +                    addr = get_peri_addr(defines, 'ADC1')
    +                else:
    +                    addr = get_peri_addr(defines, pname)
                     if addr is None:
                         continue

I have tested that the ADC works correctly on an STM32F072RB after applying this. I don't have other F0 parts to test with, but I have manually checked everything against the other F0 headers, cubedb XML files, SVD XML files, and both RM0091 and RM0360.

I can make a PR if you'd like. If there's some better way to fix this, I'd be happy to work on that too.

### Why Is The Base Address Wrong?

The headers list two base addresses: `ADC1_BASE` and `ADC_BASE`. `stm32f078xx.h` reads:

    #define ADC1_BASE             (APBPERIPH_BASE + 0x00012400UL)
    #define ADC_BASE              (APBPERIPH_BASE + 0x00012708UL)

The cubedb XML files name the peripheral `ADC`. `STM32F072R(8-B)Tx.xml` reads:

    <IP ClockEnableMode="__HAL_RCC_ADC1_CLK_ENABLE" InstanceName="ADC" Name="ADC" Version="aditf4_v1_1_Cube"/>

This causes the JSON generator to use the name `ADC_BASE` to get the peripheral address from the `#define`s in the headers. Since this is the wrong address, the code generated for the ADC peripheral does not work at all.

The value of `ADC1_BASE` is the correct base address for the peripheral. It matches the description in RM0091 (2.2.2 Memory map and register boundary addresses, Table 1. STM32F0xx peripheral register boundary addresses, pg. 49):

    | APB | 0x4001 2400 - 0x4001 27FF | 1 KB | ADC Section 13.12 on page 274 |

It matches the description in RM0360 (2.2.2 Memory map and register boundary addresses, Table 1. STM32F0x0 memory boundary addresses, pg. 40):

    | APB | 0x4001 2400 - 0x4001 27FF | 1 KB | ADC Section 12.11.11 on page 220 |

### What is `ADC_BASE` anyway?

`ADC_BASE` points at the `CCR` register, the ADC "common" configuration register (even though there's only one ADC.) The reference manuals and SVDs describe the ADC as a single register block spanning from the `ISR` through the `CCR`. They consistently refer to the `CCR` by its offset of `0x308` relative to the `ADC1_BASE` address of `0x40012400`. RM0091 reads:

    13.11.10 ADC common configuration register (ADC_CCR)
    Address offset: 0x308
    ...
    | 0x308 | ADC_CCR | ... |

RM0360 reads:

    12.11.10 ADC common configuration register (ADC_CCR)
    Address offset: 0x308
    ...
    | 0x308 | ADC_CCR | ... |

`stm32f0x2.svd` reads:

    <peripheral>
      <name>ADC</name>
      <description>Analog-to-digital converter</description>
      <groupName>ADC</groupName>
      <baseAddress>0x40012400</baseAddress>
      <addressBlock>
        <offset>0x0</offset>
        <size>0x400</size>
        <usage>registers</usage>
      </addressBlock>
    ...
        <register>
          <name>CCR</name>
          <displayName>CCR</displayName>
          <description>common configuration register</description>
          <addressOffset>0x308</addressOffset>

In constrast to the definitions in the reference manual and SVD, ST's HAL treats the ADC as two separate register blocks. One struct is mapped at `ADC1_BASE` and it contains all of the ADC registers except `CCR`. A second struct is mapped at `ADC_BASE` and it contains only the `CCR`.

In any case, there doesn't seem to be a reason to use `ADC_BASE` in here.

msamsonoff:
Fixes #156

This adds special case handling to get the correct base address for the ADC peripheral for STM32F0 devices.

I'm not sure if this is the best way to fix the problem, but it does produce correct JSON, and the ADC peripheral works correctly.

### DIFF
--- a/data/chips/STM32F030C6.json
+++ b/data/chips/STM32F030C6.json
@@ -289,7 +289,7 @@
             "peripherals": [
                 {
                     "name": "ADC",
-                    "address": 1073817352,
+                    "address": 1073816576,
                     "registers": {
                         "kind": "adc",
                         "version": "v1",
--- a/data/chips/STM32F030C8.json
+++ b/data/chips/STM32F030C8.json
@@ -289,7 +289,7 @@
             "peripherals": [
                 {
                     "name": "ADC",
-                    "address": 1073817352,
+                    "address": 1073816576,
                     "registers": {
                         "kind": "adc",
                         "version": "v1",
--- a/data/chips/STM32F030CC.json
+++ b/data/chips/STM32F030CC.json
@@ -289,7 +289,7 @@
             "peripherals": [
                 {
                     "name": "ADC",
-                    "address": 1073817352,
+                    "address": 1073816576,
                     "registers": {
                         "kind": "adc",
                         "version": "v1",
--- a/data/chips/STM32F030F4.json
+++ b/data/chips/STM32F030F4.json
@@ -289,7 +289,7 @@
             "peripherals": [
                 {
                     "name": "ADC",
-                    "address": 1073817352,
+                    "address": 1073816576,
                     "registers": {
                         "kind": "adc",
                         "version": "v1",
--- a/data/chips/STM32F030K6.json
+++ b/data/chips/STM32F030K6.json
@@ -289,7 +289,7 @@
             "peripherals": [
                 {
                     "name": "ADC",
-                    "address": 1073817352,
+                    "address": 1073816576,
                     "registers": {
                         "kind": "adc",
                         "version": "v1",
--- a/data/chips/STM32F030R8.json
+++ b/data/chips/STM32F030R8.json
@@ -289,7 +289,7 @@
             "peripherals": [
                 {
                     "name": "ADC",
-                    "address": 1073817352,
+                    "address": 1073816576,
                     "registers": {
                         "kind": "adc",
                         "version": "v1",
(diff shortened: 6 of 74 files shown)


Dirbaio (approved):
THank you!

PR #158, opened 2022-09-22
Fix DMA channels for single DAC chips

chemicstry:
Chips with multiple DACs name signals by `DACx_CHy` as by the "standard", however chips that have only a single DAC name signals by `DACy`. This PR fixes it.

### DIFF
--- a/stm32data/__main__.py
+++ b/stm32data/__main__.py
@@ -1194,6 +1194,13 @@ def parse_dma():
                         original_target_name = target_name
                         parts = target_name.split(':')
                         target_name = parts[0]
+
+                        # Chips with single DAC refer to channels by DAC1/DAC2
+                        if target_name == "DAC1":
+                            target_name = "DAC_CH1"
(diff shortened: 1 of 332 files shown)


stm32data/__main__.py
```diff
@@ -839,7 +839,13 @@ def parse_chips():
             for p in peris:
                 chs = []
                 for dma in chip_dmas:
-                    if peri_chs := dma_channels[dma]['peripherals'].get(p['name']):
+                    peri_chs = dma_channels[dma]['peripherals'].get(p['name'])
+
+                    # DAC1 is sometimes interchanged with DAC
```
chemicstry, line 844: This feels really sketchy, but I'm not sure if there is a better place to fix this.

It can't be done in `parse_dma()`, because you can't know if chip is using `DAC` or `DAC1` without multiple ifs for specific chips.

Dirbaio (approved):
Whoa, nice. Thanks for taking care of this :)

PR #159, opened 2022-11-03
Correct STM32L152CB SRAM size

dinocore1:


Dirbaio:
`data/chips/*.json` is autogenerated, could you fix the script instead so it generates the correct value?

PR #160, opened 2022-11-12
Rewrite python script in rust

GrantM11235:
This produces the exact same output as the original script, but it is much faster (even when accounting for the initial compile time!) and it should be easier to refactor in the future.

The python version "cheats" by only parsing the header files once and saving the results as json, with no way to invalidate the cached files if something changes. The new rust version is fast enough that it can parse the headers every time

- Initial run
	- Python (`rm -r sources/headers_parsed/ && time python3 -m stm32data`)
		- 57.0 seconds
	- Rust (`cargo clean && time cargo run --release`)
		- 30.0 seconds
- Second run
	- Python (doesn't re-parse headers)
		- 29.8 seconds
	- Rust (doesn't recompile)
		- 4.5 seconds

(All benchmarking was done on a beefy multi-core desktop, YMMV)

Along the way, I also found and deleted some duplicates in `header_map.yaml`


Dirbaio:
This is just :exploding_head: wow. Thanks for working on this!

I've done some changes to `[profile.release]` to reduce "build+run time after making a code change". Before it was taking 17s build + 10s run. Now it takes 2s build + 12s run.

PR #161, opened 2022-12-10
Add USB support for G0B1/G0C1 chips

stillinbeta:
This uses the https://github.com/embassy-rs/stm32-data-sources/pull/1 that adds the USB peripherals to these chips


stillinbeta:
@Dirbaio ready for another look I think!

Dirbaio:
LGTM! Thank you! :)

stillinbeta:
ah yes, I forgot about that! I'll send a followup PR and a matching
`embassy-stm32` data once I've tested everything

PR #162, opened 2022-12-17
update pwr_g0 from stm32g0b1

stillinbeta:


data/registers/pwr_g0.yaml
```diff
@@ -33,20 +33,54 @@ block/PWR:
...
+      description: Power Port A pull-up control register
       byte_offset: 32
-      fieldset: PCR
-    - name: PDCR
-      description: Power Port pull-down control register
-      array:
-        len: 6
-        stride: 8
+      fieldset: PUCR
+    - name: PDCRA
```
stillinbeta, line 40: is there a way to make these an array but have the letters still show up?

Dirbaio: Not really possible, array indices are only integers. 

There's one hack which is to keep both the PUCR array reg and the unrolled PUCRx regs. I wouldn't do it though, unrolled registers bloat the PAC and make it compile slower. The convention we're using is to arrayify everything even if it means losing letters.

stillinbeta:
The reason I don't want to just do numbers here is that PU/DCR F is a
special case, because it only has 14 registers. So it'd look weird if we
did [0,1,2,3,4,5] F

Dirbaio:
Make F have 16 bits too. Consistency is worth the slight loss of precision. The main GPIO module already does this. 

If we don't, it becomes impossible to write a sane HAL, because it causes different ports to have different Rust types, which needs horrible macro spaghetti abominations at the HAL layer to workaround.

PR #163, opened 2023-01-05
Unify and cleanup USB OTG registers

chemicstry:
This removes as many differences between FS and HS versions as possible.

The `GCCFG` register has a lot of variations between chips, because it deals with vbus, charging, etc detections. I think it can be fixed when the need arises, so I just left it as is for now


Dirbaio:
Thanks for putting in the work to cleanup this :)

PR #165, opened 2023-01-11
USB OTG: Add XCVRDLY bit

mattico:
The XCVRDLY bit is a poorly-documented feature of the Synopsis ULPI link IP which helps work around timing constraints in some ULPI PHYs. [More information here](https://github.com/stm32-rs/synopsys-usb-otg/pull/17#issue-841324871).


Dirbaio:
Thanks!

Are you going to test OTG with ULPI PHYs? Let us know if it works, it's not been tested so far :eyes:

PR #166, opened 2023-02-07
Add FMC/FSMC for the entire stm32 family

rmja:
This replaces the two current implementations which only considered stm32h7 and one of the FSMC variants.

I have been through all the datasheets and compared the variations.
The `x1` and `x3` suffix describes the number of PCR/SR/PMEM/PATT registers.

They are all as similar as possible so diff's between the versions should be easy to make. The registers, fields, and enums are sorted in the files by the appearance in the register map.


lulf (approved):
Massive thanks :rocket: 

LGTM

PR #169, opened 2023-02-08
RNG v2

lucasgranberg:
Add RNG v2.
v2 is still mostly compatible with v1 but with more registers


lucasgranberg:
Better to call it rng_v3 as it would match with rng1_v3_1?
as in:
`(".*:RNG:rng1_v3_1", ("rng", "v2", "RNG"))`

Dirbaio:
U5 and L5 also have rng_v2, could you add the right perimap entries for them?

> Better to call it rng_v3 as it would match with rng1_v3_1?

no, we do our own sequential numbering. Many peripherals already don't line up with ST's version, their number is more like a "HAL API version", not a "registers version".

lucasgranberg:
All of U5 and all of L5? Looks like some H7 also looks vaguely similar. I don't know how to validate something like that.

Dirbaio:
if you do `./d extract-all RNG` and then `grep CONDRST tmp/RNG/*` yuo get:

```
tmp/RNG/l552.yaml:    - name: CONDRST
tmp/RNG/l562.yaml:    - name: CONDRST
tmp/RNG/u5xx.yaml:    - name: CONDRST
tmp/RNG/wl5x_cm0p.yaml:    - name: CONDRST
tmp/RNG/wl5x_cm4.yaml:    - name: CONDRST
tmp/RNG/wle5.yaml:    - name: CONDRST
```

so it seems H7 is v1, assuming the SVDs are correct. if you've found a reference manual PDF that says otherwise then we should investigate

lucasgranberg:
Ok Ill go on that, u5xx + l552 + l552

Dirbaio:
these L5s are all the L5s

lucasgranberg:
allright, done

Dirbaio:
seems only u5 and wl jsons have changed, something's missing for L5?

lucasgranberg:
For some reason the U5 triggered with the custom versions at the end of perimap but the L5 version did not trigger. I moved bot L5 and U5 before the more generic rng matches and now it seems to trigger correctly.

lucasgranberg:
I got it now. U5 didn't trigger on anything, it uses rng1_v4_1 in the cubedb. Should that be added to the perimap?

Dirbaio (approved):
Looks good now! Thank you for taking the time to deal with this :)

PR #170, opened 2023-02-13
FMC->FSMC fix for F4

rmja:
cc @Dirbaio 

This is related to the regression in https://github.com/embassy-rs/embassy/pull/1213

Sometimes I wonder what exactly the naming convention for these st chips exactly is:-/


Dirbaio:
Thank you!

> Sometimes I wonder what exactly the naming convention for these st chips exactly is:-/

Conventions? Hah! :zany_face:

PR #171, opened 2023-02-13
Add LPTIM registers for STM32WB series

chrenderle:
The OR registers of LPTIM1 and LPTIM2 differ minimally. Therefore I created to separate fieldsets and enums. Hope I did everything else correct.


Dirbaio:
AFAICT the two ORs are the same. What's the difference?

chrenderle:
See the minor difference in the screenshots. LPTIM1 has innput 1 and 2 whereas LPTIM2 has only input 1.

Dirbaio:
I suspect that's a typo in the register description in LPTIM2. It makes no sense to have two bits to control the same input.

BTW, the PR got closed automatically because I've rewritten the repo history to remove the autogenerated chip json files, as the history was becoming too bloated. If you want to reopen it you'll have to rebase the changes.

PR #173, opened 2023-03-21
Add missing interrupt definitions so that `stm32-metapac` can be used in an RTIC application

noppej:
Please see this [chiptool PR](https://github.com/embassy-rs/chiptool/pull/10) for a full discussion of objectives and scope).

[x] Partially update  instructions in README.
> Some instructions were out of date, and I've updated these to the best of my understanding to date.
> My Markdown plugin made a bunch of whitespace changes. I can back these out if they are unwanted.

[x] Expose `NVIC_PRIO_BITS` in `stm32-metapac`


noppej:
@Dirbaio, I noticed that the generated `stm32-metapac` has a version number of `0.1.0`, while the `crates.io` has `1.0.0`.

I can hack this on my local tests to validate the PR changes, but is it something I should be concerned about or is it handled in the CI?

Dirbaio:
the Cargo.toml version is ... not handled at all. I hand-edited it before release :sweat_smile:

noppej:
@Dirbaio Thanks again for all your help. 

I've updated the` Cargo.toml` to use latest `chiptool`.   Also, I notice there are a bunch of clippy lint warnings. Would you like me to clear them?

Finally, the CI build fails while running `stm32-data-gen`. I can't reproduce that on my Mac. Any pointers?

Dirbaio:
> Would you like me to clear them?

why not, if you want to! 

> Update: Ignore that. I just did a fresh ./d download-all and I can reproduce ... I wonder if it has something to do with that new H5 chip thinking

yes, the new stuff in `stm32-data-sources` needs changes to stm32-data-gen. see  #175. The easiest fix is probably to change `download-all` to download one particular commit. The `stm32-data-sources` commit that works with git main is 3d60b46, the last one before adding the H5 stuff/

noppej:
It appears it is not possible to clone from a commit ... only a branch or tag. I don't have permissions to create a tag for that commit. If you can create one that would be great, or alternatively I'm happy to wait for #175 to drop first.

Dirbaio:
LGTM! I see the PR is still marked as draft. Is there something pending, or is it ready to go?

Dirbaio:
you marked it as ready literally the same second I wrote the previous message :rofl: 

Let's merge then! :rocket: 

Thank you for your work!

PR #176, opened 2023-03-25
Support multiple flash regions, including OTP

rmja:
This PR:
* Finds the best of the flash configurations instead of using "the first", preferring the largest capacity, then single bank layout over dual bank
* Append _REGION_x to banks with multiple region characteristic.
* Removes previously defined ERASE_SIZE, WRITE_SIZE and ERASE_VALUE constants
* Add OTP to generated memory.x
* Let sizes in memory.x be in kilobyte
* Makes a more stable memory lookup that takes into account the actual parsed flash size
* D1_AXIFLASH/D1_AXIICP now emits flash settings = None. I do not know whether this is correct or not.


rmja:
This is not yet ready for review...

Dirbaio:
Looks really nice! Most of the diff of generated files are bug fixes, but I found two regressions: 



- f070c6, f070f6: erase_size changed from 1k to 2k, the correct one is 1k.
- erase_size for some F1s is wrong. Many changed from 1k to 2k, but some are indeed 1k. All chips with 128k or less are 1k erase size. This is the full breakdown according to the RMs:
```
1k   Low-density    value line devices are STM32F100xx microcontrollers where the flash memory density ranges between 16 and 32 Kbytes.
1k   Medium-density value line devices are STM32F100xx microcontrollers where the flash memory density ranges between 64 and 128 Kbytes.
2k   High-density   value line devices are STM32F100xx microcontrollers where the flash memory density ranges between 256 and 512 Kbytes.
1k   Low-density       devices are STM32F101xx, STM32F102xx and STM32F103xx microcontrollers where the Flash memory density ranges between 16 and 32 Kbytes.
1k   Medium-density    devices are STM32F101xx, STM32F102xx and STM32F103xx microcontrollers where the Flash memory density ranges between 64 and 128 Kbytes.
2k   High-density      devices are STM32F101xx and STM32F103xx microcontrollers where the Flash memory density ranges between 256 and 512 Kbytes.
2k   XL-density        devices are STM32F101xx and STM32F103xx microcontrollers where the Flash memory density ranges between 768 Kbytes and 1 Mbyte.
2k   Connectivity line devices are STM32F105xx and STM32F107xx microcontrollers.
```

rmja:
@Dirbaio I was able to make this much more simple! It turns out that we have the die from the cubedb, and we can use that to uniquely identify the memory to use. This completely removes the need for regex'ing chip name patterns.

Dirbaio (approved):
LGTM, thank you!

PR #178, opened 2023-04-02
chore(docs): update readme generate section

maxekman:
Found out that `--bin` is required to run the generation commands. Got this error:

```shell
❯ cargo run --release stm32-data-gen
error: `cargo run` could not determine which binary to run. Use the `--bin` option to
specify a binary, or the `default-run` manifest key.
available binaries: stm32-data-gen, stm32-metapac-gen
```


Dirbaio (approved):
Thanks!

PR #180, opened 2023-04-04
feat(sai): add complete peripherals

maxekman:
Adding all versions of SAI peripheral I could find. The existing partial implementation of `sai1_v1_1` is in fact a "v2" so it has been updated with better docs and renamed.

- [x] Add SAI1
- [x] Add SAI2-4 depending on devices
- [x] Remove "useless enums"


Dirbaio:
looks great!

Could you remove the "useless" enums? see https://github.com/embassy-rs/stm32-data#register-cleanup

maxekman:
@Dirbaio Thanks, absolutely! Need to figure out how to make the CI pass too, maybe you have some advice? I'm new to this project.

Dirbaio:
make sure to name the main block `SAI`. You have `SAI1`.
It has to match what you put in `perimap`. For example `("sai", "v1", "SAI")` means "block `SAI` in `sai_v1.yaml`"

maxekman:
(I'll keep pushing fixes one by one, will squash in the end.)

maxekman:
@Dirbaio Please have a look and see if you think this is good enough. I have to admit that I haven't worked on this in a while, but just rebased it. Would be nice to wrap up if you are happy with it.

maxekman:
You should probably also add the SAI peripherals to the register versions sheet: https://docs.google.com/spreadsheets/d/1-R-AjYrMLL2_623G-AFN2A9THMf8FFMpFD4Kq-owPmI/edit#gid=1972450814

Dirbaio:
Thank you!

PR #181, opened 2023-04-06
Feat/add uid regs

ericyanush:
This PR adds support for the Unique ID regs on all chips.

Based on the zephyrproject-rtos/zephyr#14208 PR, this is supported on all STM32 families/parts, and this seems to be confirmed by the presence of the UID_BASE definition in all part header files.

I originally implemented ([in this commit](https://github.com/embassy-rs/stm32-data/pull/181/commits/910246a88c159539fee377d98cf43d6ceb2e8f66)) the register mapping based on the "breakdown" specified in the L4x1 ref manual, which identifies the wafer x/y coordinates, wafer and lot numbers, however, this isn't defined for all families. Based on a preliminary investigation it was defined starting on the F0 line, and likely all newer parts, but I don't have cycles to verify that as this point. I also question the usefulness of having the lot/wafer/coord breakdown.

### DIFF
--- a/data/registers/uid_v1.yaml
+++ b/data/registers/uid_v1.yaml
@@ -1,0 +1,45 @@
+---
+block/UID:
+  description: Device Factory programmed 96-bit unique device identifier
+  items:
+    - name: UID0
+      description: Factory programmed 96-bit unique device identifier word 0
+      byte_offset: 0
+      access: Read
+      fieldset: UID0
+    - name: UID1
+      description: Factory programmed 96-bit unique device identifier word 1
+      byte_offset: 0x04
+      access: Read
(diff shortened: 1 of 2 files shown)


data/registers/uid_v1.yaml
```diff
@@ -0,0 +1,18 @@
+---
+block/UID:
+  description: Device Factory programmed 96-bit unique device identifier
+  items:
+    - name: UID0
```
Dirbaio, line 5: Could you make this an array?

ericyanush: Done!

Dirbaio (approved):
Thank you!

PR #185, opened 2023-04-10
i2s: add pins

xoviat:


xoviat:
@Dirbaio Is there anything else to do to have this merged?

Dirbaio:
LGTM, except: could you add `I2S_` prefix to the I2S pin signals? so that it's `I2S_CK` instead of just `CK`. This will make the naming less confusing.

Dirbaio (approved):
Thank you!

PR #188, opened 2023-04-30
add hrtim

xoviat:


Dirbaio:
Could you clean up the register yaml according to https://github.com/embassy-rs/stm32-data#register-cleanup ?

xoviat:
Moving this back to draft as there are more enums to cleanup.

xoviat:
I think the enums are sufficiently cleaned up at this point. Let me know if there is anything else to do.

Dirbaio:
there's a few 
- chips.rs points to `block/HRTIM`, but that doesn't exist. There should be one main register block, which can then contain the others.
- G4 has a different HRTIM, with 6 timers and adds other stuff.

there's still a few enums that could be removed
- there's a few enums that are 0=inactive, 1=active, these could be removed as well
- enum/EVENT, same
- enum/SDTFx and enum/SDTRx are the same

and there's still a lot of duplications in registers/fields.
- TIM[ABCDE] could be merged and made into an array, they're identical.
- compare 1-4 could be made into an array
- output 1-2 could be made into an array
- fields for "extenral event X" capture/latch/filter could be made into arrays
- fault 1-5, same

The reason all this cleanup is needed is that otherwise it's impossible to write a HAL on top of it, without tons of macros.

xoviat:
TODO: `fieldset/RSTXR`

xoviat:
end 1014

Dirbaio:
let's gooo

PR #189, opened 2023-05-01
adding STM32G4 flash peripheral

dognotdog:
Compared to existing G0 flash peripheral, and differences check out on first look, though did not go through the datasheet with a fine comb.


Dirbaio:
Thanks!

PR #190, opened 2023-05-08
add enums to STM32G4 Flash/RCC register YAMLs

dognotdog:
Modeled enums on STM32G0 definitions for RCC/FLASH peripherals of STM32G4. Checked enums against RM0440 Rev 7.


Dirbaio (approved):
Thanks!

PR #193, opened 2023-06-01
can: skip duplicate interrupt in f107

xoviat:
This skips interrupts with a peripheral named can. This peripheral does not exist, but causes later duplicate interrupts with the correct peripheral to be skipped. If an IP file contains only interrupts with an incorrect CAN peripheral, this could cause those interrupts to be removed. However, I am not aware of such a file and am not sure how I would identify it.

### DIFF
--- a/.vscode/.gitignore
+++ b/.vscode/.gitignore
@@ -1,0 +1,2 @@
+launch.json
+tasks.json
--- a/stm32-data-gen/src/chips.rs
+++ b/stm32-data-gen/src/chips.rs
@@ -935,13 +935,40 @@ fn process_core(
         }
 
         if let Some(peri_irqs) = chip_irqs.get(&pname) {
+            use stm32_data_serde::chip::core::peripheral::Interrupt;
+
             //filter by available, because some are conditioned on <Die>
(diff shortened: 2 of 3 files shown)


xoviat:
There's some nonsense going on with `stm32f398ve` that I need to solve as well.

Dirbaio:
Checking the diff of generated jsons, it looks okay for F105 and F107, but in F103 it's removing all CAN interrupts.

xoviat:
What appears to be happening is the interrupt definition file is a list of "potential interrupts" that are later selected in the chip definition file. This is not possible to represent with the current model, because the interrupts are first stored in an a hashmap that has the interrupt name as the key, meaning that if an interrupt could belong to two different peripherals, it is eliminated from the latter peripheral. 

What do you think would be a good solution to this? I'm sure it was done this way for a reason, but the current hashmap does not appear to work with can in all cases.

xoviat:
@Dirbaio These changes should address the issue. Basically, I passed through duplicate interrupts and deduplicate them in the chips code. By inspection, only CAN interrupts had this problem.

Dirbaio:
Thanks! :)

PR #198, opened 2023-06-19
Feature/stm32g crs

kevswims:
Adds the CRS peripheral to the STM32 G0 and G4. Previously this was only used on the L0.


stm32-data-gen/src/chips.rs
```diff
@@ -284,7 +284,9 @@ impl PeriMatcher {
             ("STM32H50.*:EXTI:.*", ("exti", "h50", "EXTI")),
             ("STM32H5.*:EXTI:.*", ("exti", "h5", "EXTI")),
             (".*:EXTI:.*", ("exti", "v1", "EXTI")),
-            ("STM32L0.*:CRS:.*", ("crs", "l0", "CRS")),
+            ("STM32L0.*:CRS:.*", ("crs", "v1", "CRS")),
```
kevswims, line 287: Note: I couldn't find an actual silicon version for these. Is v1 the correct thing to label this?

stm32-data-gen/src/chips.rs
```diff
@@ -284,7 +284,9 @@ impl PeriMatcher {
...
+            ("STM32G0.*:CRS:.*", ("crs", "v1", "CRS")),
```
kevswims, line 288: I'm not quite sure if all G0 chips have the CRS peripheral. Is there a better way to constrain this down or does it matter?

stm32-data-gen/src/chips.rs
Dirbaio: yeah, that's fine. Most v1/v2/v3... is invented by us in roughly chip release chronological order.

Dirbaio:
Thank you!

PR #201, opened 2023-06-28
G0 SYSCFG: fix PA11/PA12 RMP definitions

cbiffle:
It looks like the G0 file here picked up the F0-style "one remapping bit for both pins" field definition. The G0 series actually has C0-style dual remapping bits.

### DIFF
--- a/data/registers/syscfg_g0.yaml
+++ b/data/registers/syscfg_g0.yaml
@@ -186,8 +186,12 @@ fieldset/CFGR1:
       bit_offset: 0
       bit_size: 2
       enum: MEM_MODE
-    - name: PA11_PA12_RMP
-      description: PA11 and PA12 remapping bit.
+    - name: PA11_RMP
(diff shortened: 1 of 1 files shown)


data/registers/syscfg_g0.yaml
```diff
@@ -186,8 +186,12 @@ fieldset/CFGR1:
       bit_offset: 0
       bit_size: 2
       enum: MEM_MODE
-    - name: PA11_PA12_RMP
-      description: PA11 and PA12 remapping bit.
+    - name: PA11_RMP
+      description: "PA11 pin remapping\r This bit is set and cleared by software. When set, it remaps the PA11 pin to operate as PA9 GPIO port, instead as PA11 GPIO port."
+      bit_offset: 4
```
Dirbaio, line 191: this should be 3

Dirbaio:
huh, nice find. The mistake comes from ST's SVD it seems :upside_down_face:

data/registers/syscfg_g0.yaml
cbiffle: Fixed!

PR #205, opened 2023-06-28
G0/C0: fix SYSCFG.CFGR1.MEM_MODE definition

cbiffle:
- Added docs from reference manual / fixed existing partial copy of reference manual docs
- Renamed System Flash case to SYSTEM_FLASH because FLASH was ambiguous (you really don't want to activate the System Flash when you wanted Main Flash, where your program lives!)
- Added Main Flash case.

Stylistically: note that main flash is actually `x0`. So the value I've given is valid for writes, but on read `== MAIN_FLASH` here could be misleading. Not sure how you prefer to handle cases like that.

### DIFF
--- a/data/registers/syscfg_c0.yaml
+++ b/data/registers/syscfg_c0.yaml
@@ -123,7 +123,7 @@ fieldset/CFGR1:
   description: configuration register 1
   fields:
     - name: MEM_MODE
-      description: "Memory mapping selection bits\r This bitfield controlled by software selects the memory internally mapped at the address 0x0000 0000. Its reset value is determined by the boot mode configuration. Refer to for more details.\r x0: Main Flash memory"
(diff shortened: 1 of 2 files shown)


Dirbaio:
> Stylistically: note that main flash is actually x0. So the value I've given is valid for writes, but on read == MAIN_FLASH here could be misleading. Not sure how you prefer to handle cases like that.

i've done something like `MAIN_FLASH_ALT`, value of `2` in the past when encountering this, so when writing a `match` you see both cases. Not great, but I think better than not having it.

cbiffle:
Can do -- though note that currently when writing a match you'll see all numbers from 0 to 255 due to that issue we're discussing.

Dirbaio:
true :upside_down_face:. At least you'll see it in the IDE autocomplete though.

Issue #206, opened 2023-06-28
Modeling of AFL/AFH registers presents a bit of a footgun


cbiffle:
I've now ported a nontrivial firmware application over to stm32-metapac and I'm delighted with the results. It has enabled me to extend my G0 prototype to support C011 with very few code changes. So, thank you for all the hard work there.

I ran into something interesting during the porting process. First, I kept getting zero-byte binaries, but this was my own mistake.  (In case anyone's reading this with that problem: make sure you've specified the `rt` and `pac` features and included the `-Tlink.x` linker arg in your target triple. And not, say, in a different but unrelated target triple you copied from another project. Learn from my mistakes.)

Then I started getting very small binaries -- like 5 kiB instead of the ~18 kiB I was expecting. This is almost always a sign that rustc has decided that some code has become an unconditional panic or UB and eliminated the rest of the program.

After spending some time hunting and reading the machine code, the problem turned out to be my use of the AFRH register in the GPIO block. Previously, using the traditional PAC, I had

```rust
    p.GPIOA.afrh.write(|w| {
        w.afsel9().af1(); // USART1_TX
        w.afsel10().af1(); // USART1_RX
        w
    });
```

During the port to the metapac, I rewrote this code to

```rust
    device::GPIOA.afr(1).write(|w| {
        w.set_afr(9, 1); // USART1_TX
        w.set_afr(10, 1); // USART1_RX
    });
```

(This reflects the discovery that the registers described in the reference manual as AFRL / AFRH are modeled in the metapac as a two-element array -- which I'm not totally convinced about, but I can see the advantages for HALs!)

This code was compiling into an unconditional program-ending panic. Perhaps you can see the problem already. :-) If, like me, it isn't obvious, the answer is: the `AFR` field arrays in each of the two AFRx registers are treated as 8-element arrays. This means to set (say) AFRH9, you need to `set_afr(1, 1)` instead of `set_afr(9, 1)`.

Given that the fields in the AFRH register are numbered from 8 in the reference manual and all other SDKs, this seems like it's going to bite other people in the future. I've altered my code to use `set_afr(9 - 8, 1)` to try and remind me what's going on in the future.

I wonder if you can think of a way to address? Since the setters already contain an index-checking assert, I suppose they could insist that the accesses start at 8 and subtract 8, but... even though that would have fixed my bug, it seems really weird and inconsistent.

[...]

PR #207, opened 2023-06-28
gpio_v2: expand docs on AFRx register pair to describe indexing.

cbiffle:
In an attempt to keep other people from making the mistake I described in #206, I've expanded the docs to explain how indexing of the fields works (because I don't expect users to read and understand the YAML).

These docs don't meet my normal standards for Rustdoc: they don't include a code example, they don't link the types involved using Markdown references, etc. However, it appears that the description field in the YAML is intended to be language-neutral; I don't think good API documentation can be language neutral, so you may want a way of adding language-specific documentation in the future.

For now, I have described it in very generic terms that apply to any language using 0-based array indexing, which does not cover all languages used in embedded (e.g. Lua, or in some cases Ada), but does cover C and Rust and Python.

### DIFF
--- a/data/registers/gpio_v2.yaml
+++ b/data/registers/gpio_v2.yaml
@@ -37,7 +37,7 @@ block/GPIO:
       byte_offset: 28
       fieldset: LCKR
     - name: AFR
-      description: "GPIO alternate function register (low, high)"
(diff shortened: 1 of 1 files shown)


cbiffle:
Wow, thanks, that was fast -- turns out there's another issue where the current docs are misleading that I was trying to cram in here. I will open a second PR. :-)

PR #209, opened 2023-06-30
update RNG registers and mapping

JcBernack:
It seems to me that the "IP Version" for the RNG peripheral is not really reliable. I changed the peripheral mapping to the chip family/names instead.

From why I see there are three different variants out there:
- v1: default
- v2: adds the HTCR register and conditioning on init
- v3: adds auto-reset bit ARDIS


stm32-data-gen/src/chips.rs
```diff
@@ -130,12 +130,11 @@ impl PeriMatcher {
...
+            ("STM32[HU]5.*:RNG:.*", ("rng", "v3", "RNG")), // adds auto-reset bit ARDIS
             ("STM32L5.*:RNG:.*", ("rng", "v2", "RNG")),
-            (".*:RNG:rng1_v1_1", ("rng", "v1", "RNG")),
-            (".*:RNG:rng1_v2_0", ("rng", "v1", "RNG")),
-            (".*:RNG:rng1_v2_1", ("rng", "v1", "RNG")),
-            (".*:RNG:rng1_v3_1", ("rng", "v2", "RNG")),
-            (".*:RNG:rng1_v4_1", ("rng", "v2", "RNG")),
+            ("STM32L4[PQ]5.*:RNG:.*", ("rng", "v2", "RNG")),
+            ("STM32WL.*:RNG:.*", ("rng", "v2", "RNG")),
+            (".*:RNG:.*", ("rng", "v1", "RNG")), // use rng_v1 for everything else
```
Dirbaio, line 137: this is somewhat dangerous, it makes it easy to end up with wrong regs when adding new chips in the future. it's better to list all families explicitly, so newer chips get no regs instead of wrong regs.

Dirbaio:
Thanks!

PR #212, opened 2023-07-01
Add LPTIM registers for STM32WB series

chrenderle:
I had previously created a pull request  (#171) but didn't get the rebase to work so I discarded the changes and created a new commit. Hope this is okay

### DIFF
--- a/data/registers/lptim_v1c.yaml
+++ b/data/registers/lptim_v1c.yaml
@@ -1,0 +1,266 @@
+---
+block/LPTIM:
+  description: Low power timer
+  items:
+    - name: ISR
+      description: Interrupt and Status Register
+      byte_offset: 0
+      access: Read
+      fieldset: ISR
+    - name: ICR
+      description: Interrupt Clear Register
+      byte_offset: 4
+      access: Write
+      fieldset: ICR
+    - name: IER
+      description: Interrupt Enable Register
+      byte_offset: 8
+      fieldset: IER
+    - name: CFGR
+      description: Configuration Register
+      byte_offset: 12
+      fieldset: CFGR
+    - name: CR
+      description: Control Register
+      byte_offset: 16
+      fieldset: CR
+    - name: CMP
+      description: Compare Register
+      byte_offset: 20
+      fieldset: CMP
+    - name: ARR
+      description: Autoreload Register
+      byte_offset: 24
+      fieldset: ARR
+    - name: CNT
+      description: Counter Register
+      byte_offset: 28
+      access: Read
+      fieldset: CNT
+    - name: OR
+      description: Option Register
+      byte_offset: 32
+      fieldset: LPTIM_OR
+fieldset/ARR:
+  description: Autoreload Register
+  fields:
+    - name: ARR
+      description: Auto reload value
+      bit_offset: 0
+      bit_size: 16
+fieldset/CFGR:
+  description: Configuration Register
+  fields:
+    - name: CKSEL
+      description: Clock selector
+      bit_offset: 0
+      bit_size: 1
+    - name: CKPOL
+      description: Clock Polarity
+      bit_offset: 1
+      bit_size: 2
+    - name: CKFLT
+      description: Configurable digital filter for external clock
+      bit_offset: 3
+      bit_size: 2
+    - name: TRGFLT
+      description: Configurable digital filter for trigger
+      bit_offset: 6
+      bit_size: 2
+    - name: PRESC
+      description: Clock prescaler
+      bit_offset: 9
+      bit_size: 3
+    - name: TRIGSEL
+      description: Trigger selector
+      bit_offset: 13
+      bit_size: 3
+    - name: TRIGEN
+      description: Trigger enable and polarity
+      bit_offset: 17
+      bit_size: 2
+    - name: TIMOUT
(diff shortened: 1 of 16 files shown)


xoviat:
Is this IP present in any other chips?

chrenderle:
> Is this IP present in any other chips?

According to the table stm32 register versions, no.
But it also says that the LPTIM versions need more checking so I'll check the last few chip series.

chrenderle:
Okay so I've checked all Reference Manuals from STM32F1 to STM32H5 and there were none which had exactly the same registers. Even the RM 0471 (STM32WB50CG/30CE), RM 0473 (STM32WB15CC) and RM 0478 (STM32WB10CC) have some little differences (less LPTIM_OR registers or difference in LPTIM1_OR fields). 

In general there are many differences in the series which have a LPTIM and it seems to me that the versioning in the stm32 register versions table is insufficient. It only has v1, v1a, v1b, v1c, v1d and v2. So far I've only compared the Reference Manuals to the STM32WB55xx and WB35xx series (RM 0434) and didn't check all the details but my feeling is that there are more differences than are currently reflected in the table. 

If I find the time, I'll compare all the Reference Manuals and will create a more accurate versioning or at least a table with all the different occurrences.

chrenderle:
So, I've now gone through all Reference Manuals again and wrote the yaml file for all different variations. In total there are 12 when you consider every new or missing field (1 bit) an extra variation but not the descriptions for the fields. For example in RM 0401 the LPTIM1_OR vs in RM 0434 the LPTIM1_OR 
image
image I didn't consider a variation. Only if the OR_1 was missing for example would it be a new variation.

I still need to check everything again and find a proper naming scheme and then I'll add it to this PR. But since this can take some time and I don't know when I'll find the time again, I'll upload my files on my fork so somebody else could continue it (files are here https://github.com/chrenderle/stm32-data/commit/be12079701eb0af3b98cddf56322fc9fab02150b). The main part is already done.

Dirbaio:
STM32 timers are giant loads of fun.

In non-LP timers, the registers are the same except these magic "option registers" that have special cases to connect some inputs. What I did was to ignore the option registers, so all timers could be the same version. I think it's the same for LPTIM, so I would do the same.

The reason is it's otherwise unmaintainable. Copypasting all the register yaml just to edit the option bits means the regs are duplicated too many times. Also, these bits are different between different instances of timers in the same chip, which would mean  they'd need different structs, which would make it hard to write a HAL that works with any timer dynamically.

The downside is we can't use these bits, but they're very rarely used anyway. If we wanted to add them, we could add an extra "TIM_OR" peripheral that contains just these bits, so that the main TIMx can keep being the same. I did this with "USB" vs "USBRAM" and it worked quite well.

Also, LPTIM is really another version of TIM. The basic structure of the regs is the same. Perhaps it makes sense to put it in `timer_v1.yaml` as another block, instead of making separate `lptim_xx.yaml` files. This happened with UART vs LPUART, in the end it made more sense to merge them.

chrenderle:
Okay so here are all the differences with the numbers on how many of the variances are included by them:
- LPTIM_CR
	- has 5 fields -> 9/12
	- has 3 fields -> 3/12
- LPTIM_RCR and LPTIM_CCMR1
	- with RCR -> 1/12
	- with RCR and CCMR1/CCMR2 -> 2/12
	- without both -> 9/12
- LPTIM_ISR alternates -> 1/12
- LPTIM_ISR
	- has 7 fiels -> 9/12
	- has 9 fields -> 1
	- has 10 fields -> 1
	- has 12 fields -> 1
- LPTIM_CCR1 has/hasn't -> 2/12
- LPTIM_CFGR with/without TRIGSEL -> 1/12
- LPTIM_OPTR has/hasn't -> 1/12
- LPTIM_CFGR2
	- has 4 fields -> 2/12
	- hasn't -> 9/12
	- has 2 fields -> 1/12
- LPTIM3_CFGR2 has/hasn't -> 1/12

Before your message I would have proposed the following variations:
- variation a -> 6/12
	- LPTIM_CR has 5 fields
	- without LPTIM_RCR
- variation b -> 3/12
	- LPTIM_CR has 3 fields
	- without LPTIM_RCR
- variation c -> 2/12
	- LPTIM_CR has 5 fields
	- with LPTIM_RCR
- variation d -> 1/12
	- LPTIM_ISR alternate between input and output
Variation a and c could be included with "extends" in variation b in 1 file. Variation d is too different with the alternation to include it somehow.

The rest of the differences could be added later in an extra file if someone needs them. This would be the following differences:
- LPTIM_OR -> didn't count, i think 3-4
- LPTIM_CFGR2
	- has 4 fields -> 2/12
	- has 2 fields -> 1/12
- LPTIM_OPTR -> 1/12
- LPTIM3_CFGR2 -> 1/12
- LPTIM_CFGR only TRIGSEL -> 1/12
- LPTIM_CCR1 -> 2/12
- LPTIM_ISR
	- 9 fields -> 1/12
	- 10 fields -> 1/12
	- 12 fields -> 1/12

But in the first step I think it makes sense to just create one basic version which all RMs (except the alternating one) can use and if someone needs these extra registers for a specific feature we could add more variations or add these as extra files just for this register. As you said we could include them in timer_v1.yaml.

PR #213, opened 2023-07-01
hrtim: add common registers and v2

### LINKED ISSUE #210: hrtim v1 "common functions" registers are missing
These were apparently not included in the initial scrape or were removed during cleanup. The registers are from offset 0x380 - 0x3FF.

xoviat:
this pr *is* complete and aims to add the rest of the common registers and hrtim v2

v1.1 is binary compatible, but accessing the DLL is UB. However, that can be handled in the HAL.

closes #210

### DIFF
--- a/data/registers/hrtim_v1.yaml
+++ b/data/registers/hrtim_v1.yaml
@@ -111,10 +111,50 @@ block/HRTIM:
       description: "High Resolution Timer: External Event Control Register 3"
       byte_offset: 0x3b8
       fieldset: HRTIM_EECR3
+    - name: ADC1R
+      description: "High Resolution Timer: ADC Trigger [1, 3] Register"
+      byte_offset: 0x3bc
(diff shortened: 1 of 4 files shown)


Dirbaio:
The diff between hrtim v1 and v2 looks a bit suspicious. Are you sure it's correct? For example:

screenshot-2023-07-07_17-33-49

xoviat:
I assume you are referring to ADCTC3? On hrtim v1, it's bit 11, 15, 19, 24, and 29 (RM0364 p786-787). On hrtim v2, it's bit 29 and 15 (because the order is reversed) (RM0440 p1047-1048).  Or to what are you referring?

Dirbaio:
Okay! If it's really like that then LGTM. Thanks! It just looked super weird, the bit numbers are like entirely random. Weird things ST does :shrug:

PR #215, opened 2023-07-04
hsem

xoviat:


data/registers/hsem_wb55.yaml
```diff
@@ -0,0 +1,220 @@
...
+        stride: 4
+      byte_offset: 128
+      access: Read
+      fieldset: RLR
+    - name: IER
+      description: HSEM Interrupt enable register
+      byte_offset: 256
+      fieldset: IER
+      array:
+        offsets:
```
Dirbaio, line 25: these could be len=2, stride=16

data/registers/hsem_wl5.yaml
```diff
@@ -0,0 +1,269 @@
...
+      byte_offset: 324
+      fieldset: KEYR
+fieldset/ICR:
+  description: HSEM Interrupt clear register
+  fields:
+    - name: ISC
+      description: Interrupt(N) semaphore n clear bit
+      bit_offset: 0
+      bit_size: 1
+      enum_read: C1ICR_ISC0R
```
Dirbaio, line 70: enum_read/enum_write is removed in the latest chiptool, can you remove them?

if we want to keep the enum it's just `enum` now. anyway, the enums here are "0 = not pending, 1 = pending", or "write 1 to clear", which are enums we typically remove.

taunusflieger:
This looks like the PR is based on an old version (v1.3 ??) of the svd file which is missing the second IRQ bank (`HSEM_C2IER, HSEM_C2ICR,HSEM_C2ISR, HSEM_C2MISR`) for the second core. Newer versions of the svd have also a name change to comply with the rm0399 spec (i.e. `MASTERID` -> `COREID`). I'm interested in getting this PR in as I need HSEM support. 
@xoviat are you still working on this - can I help?

Dirbaio:
it seems abandoned, I'd recommend you take the work here, do the fixes you mention and open a new PR.

Dirbaio:
done in #450

PR #217, opened 2023-07-05
extract-all: Distinguish missing and failure

mkj:
New svd files sometimes fail with other errors like bad enum widths, so distinguish between that and a missing peripheral.


Dirbaio:
Nice! thanks

PR #220, opened 2023-07-08
i3c: add

xoviat:


Dirbaio:
- Can you remove all "useless" enums? All "0=disabled, 1=enabled", all "write 1 to do action", all "write 1 to clear".
- Can you add enum names instead of the autogenerated `B_0x1`?

Dirbaio:
done in #415

PR #224, opened 2023-07-25
Add aditf5_v3_1 ADC

JuliDi:
ADC is currently unavailable for some STM32H7 chips like the H7A3. The registers seem to be the same across RM0455 and RM0399 and the ADC is working on a Nucleo H7A3ZI-Q board.

This PR therefore adds aditf5_v3_1 as a v4 ADC.


Dirbaio (approved):
Thanks!

Issue #225, opened 2023-07-29
STM32U5 is compatible with CRS v1 but CRS registers are not generated for U5 chips


tyler-gilbert:
I am new to embassy, but from what I can tell, there just needs to be another line added to `chips.rs` here specifying to use `("STM32G4.*:CRS:.*", ("crs", "v1", "CRS"))`.

```
            ("STM32G0B1.*:CRS:.*", ("crs", "v1", "CRS")),
            ("STM32G0C1.*:CRS:.*", ("crs", "v1", "CRS")),
            ("STM32G4.*:CRS:.*", ("crs", "v1", "CRS")),
```

Does that seem like the right solution?

Dirbaio:
If the registers are indeed identical yes, that's the right fix. Can you send a PR?

tyler-gilbert:
I will send a PR.

PR #226, opened 2023-07-30
Fix Issue #225. Add CRS v1 to STM32U5 chips.

tyler-gilbert:
The STM32U5 uses the same CRS peripheral as the L0, G0, and G4 products. This adds the association of the CRS v1 peripheral to the U5 CRS registers.


Dirbaio (approved):
thanks!

PR #232, opened 2023-08-06
Add adc common f3

dreilly1982:
Add mappings for ADC_Common registers and vrefintcal.


xoviat:
Thunderdome: can you merge the ended/notended enums and remove the enabled/disabled enums? do you need help with that?

xoviat:
I've taken care of it.

PR #235, opened 2023-08-07
rework f3 series rcc take 2

dreilly1982:
Reworked so f3 series (except STM32F37xx) will now not have ADC periphs RCC mapped, will handle that in the hal code.  Also added registers for STM32F37xx ADC and RCC.  Still must do SDADC.


dreilly1982:
The DBGMCUEN register is only documented on the F37 chip, and is included in the rcc_v3_v2.yaml If this is present in all chips but undocumented, I can easily put it back in.

dreilly1982:
ok, added the DBGMCU registers back in

dreilly1982:
found bug, marking as draft

dreilly1982:
Found the bug, and fixed it.

Dirbaio (approved):
oooh, this is great. It gets *so many* new ADC RCC bits. Thanks for working on this!

PR #237, opened 2023-08-18
split H7 flash for stm32h7a3/b3/b0 chips

randomplum:
Some chips in the H7 family (stm32h7a3/b3/b0) have a different flash peripheral (different CR register and different latency/wait state configurations).


Dirbaio (approved):
thanks!

PR #241, opened 2023-09-01
Add UCPD

ExplodingWaffle:
Marking as draft, would appreciate review of my TODOs and CFGR3 stuff.

Some general UCPD notes:

The following parts have UCPD. I think they're all the same, apart from these differences:

[g0b1 g0c1 g071 g081][g0x1] - CR has DBATT (and CC[1, 2]VCONNEN, only in the SVDs for g0b1 and g0c1) , CFGR3 has 3 trims. Registers IPVER, IPID and MID are present in g071 and g081 svd, but not in RM since we probably aren't supposed to know about these.
[g4a1 g431 g441 g471 g491 g473 g483 g474 g484][g4] - offset 0x008 (CFGR3) is reserved
[h562 h563 h573][h562 and h5x3] and [u575 u585 u595 u5A5 u599 u5A9][u5] - CR has CC[1, 2]VCONNEN in SVDs but not RM, CFGR2 has RXAFILTEN (not in SVDs), CFGR3 has 2 trims
[l552 l562][l5x2] - CR has DBATT, CFGR3 has 3 trims

I'm not sure what CCxVCONNEN is for, if anything. (It kinda sounds like this would enable some kind of internal VCONN switch or GPIO AF, but this isn't a documented feature.)

The U5 and H5 SVDs have enumeratedValues, so I used U5 (H5 makes chiptool upset). not 100% sure if this was a good idea.

CFGR3 smells like a bit of a hack (under "UCPD implementation" in the RMs, "fully automatic trimming" is described as a "feature" which is available in some parts but not others. There's a slightly suspicious [errata](https://www.st.com/resource/en/errata_sheet/es0418-stm32g071xx-device-errata-stmicroelectronics.pdf) on g071 and g081 but I think it's unrelated).
G0 and L5 have this "feature" but still have a CFGR3 register, containing TRIM[1, 2]_NG_CC[3A0, 1A5, RPD] fields for the 3 different pull-up values for each CC line.
H5 does not have the feature and U5 has it only on some parts. The H5 and U5 RMs describe a procedure for grabbing some magic values out of "non-volatile memory" (which is in the OTP on U5, and... part RESERVED just after FLASH, part RO memory right with UID on H5??? super cursed!!!), then writing them to TRIM_CC[1, 2]\_[RP, RD] (for the CC pull-up and the pull-down), including every time you change the value of the pull-up. In the U5 SVD, these fields are typo'd to have the same names they do on G0/L5- in the H5 SVD, they're straight up missing. Also, the U5 (and H5 but not applicable there) RM says "The register is reserved (must not be written) for devices that support the fully automatic
trimming."

All of these have the same version "ucpd_v1_0_Cube" in the cubedb xmls. 🙄 

Does this CFGR3 Cursedness qualify as a seperate version of the peripheral? I imagine that could be quite annoying (especially with U5).

Side-note: this has been fairly easy and pleasant, nice job on the project and the tooling :) if anyone else wants to add a peripheral, go ahead and try, it's pretty easy (though I imagine difficulty increases with more MCUs and peripheral versions to check). The only thing I tripped up on- It turns out you need to rename the block in the yml you create to e.g UCPD instead of UCPD1... which makes me think this is what [this was supposed to say](https://github.com/embassy-rs/stm32-data/blob/9a61a1f090462df8bd1751f89951f04934fdceb3/README.md?plain=1#L99C22-L99C22) ;). And I guess also metapac taking a while to compile but it's not the end of the world and probably more to do with my crap PC.

[g0x1]: https://www.st.com/resource/en/reference_manual/rm0444-stm32g0x1-advanced-armbased-32bit-mcus-stmicroelectronics.pdf
[g4]: https://www.st.com/resource/en/reference_manual/rm0440-stm32g4-series-advanced-armbased-32bit-mcus-stmicroelectronics.pdf
[h562 and h5x3]: https://www.st.com/resource/en/reference_manual/rm0481-stm32h563h573-and-stm32h562-armbased-32bit-mcus-stmicroelectronics.pdf
[u5]: https://www.st.com/resource/en/reference_manual/rm0456-stm32u5-series-armbased-32bit-mcus-stmicroelectronics.pdf
[l5x2]: https://www.st.com/resource/en/reference_manual/rm0438-stm32l552xx-and-stm32l562xx-advanced-armbased-32bit-mcus-stmicroelectronics.pdf

### DIFF
--- a/data/registers/ucpd_v1.yaml
+++ b/data/registers/ucpd_v1.yaml
@@ -59,7 +59,21 @@ block/UCPD:
       description: "Rx ordered set extension register 2 \t"
       byte_offset: 56
       fieldset: RX_ORDEXTR2
-      # TODO: IPVER, IPID and MID from g0[7, 8]1 svds? do we care?
+    - name: IPVER
+      description: UCPD IP ID register
+      byte_offset: 1012
+      access: Read
+      fieldset: IPVER
+    - name: IPID
+      description: UCPD IP ID register
+      byte_offset: 1016
+      access: Read
+      fieldset: IPID
+    - name: MID
+      description: UCPD IP ID register
+      byte_offset: 1020
+      access: Read
+      fieldset: MID
 fieldset/CFGR1:
   description: "configuration register 1 "
   fields:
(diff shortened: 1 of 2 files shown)


ExplodingWaffle:
Now tested on hardware (g071rb) and seems to work fine 👍 caught and fixed a couple things, this should be good to go now.
(I've made CFGR3 match H5/U5, there's no good reason to access it on anything else anyway)

PR #254, opened 2023-09-13
Handle "_C" pins

JuliDi:
This PR part of the "clean_pin" logic that removed all "_C" pins present on most H7 chips.
It also changes the `Pin` type to use a `String` to store the actual pin (e.g. PA5) to be able to accommodate the _C pins (e.g. PB2_C).
Since using strings breaks the sorting (alphabetic sorting != numeric sorting), a custom implementation of the `Ord` trait was necessary for the `Pin` struct.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -620,6 +620,8 @@ version = "0.1.0"
 dependencies = [
  "itertools",
  "rayon",
+ "ref_thread_local",
+ "regex",
  "serde",
  "serde_json",
 ]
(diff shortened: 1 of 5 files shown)


Dirbaio:
can you rebase so this picks up the features in #255?

it allows using the generated `stm32-metapac` in the corresponding `embassy-stm32` PR https://github.com/embassy-rs/embassy/pull/1895 so we can see that everything works end-to-end. 

to do that, please do [these](https://github.com/embassy-rs/embassy/pull/1889/files) changes to the embassy PR (changing the job ID of course)

let's try how it works with a real contribution :) hopefully it's not too burdensome, if it is we'll have to rethink the process a bit.

JuliDi:
Sure, I can do it later today and report back how it worked out. 
Seems like a good idea being able to do this kind of end to end test.

JuliDi:
@Dirbaio could you have a look to see whether it worked?

JuliDi:
On this side everything seems fine. It's only the HAL PR that has some unresolved issues.

PR #256, opened 2023-09-14
Pwr l0 and some cleanup

xoviat:


data/registers/pwr_l0.yaml
```diff
@@ -228,24 +209,6 @@ enum/VOS:
...
-      value: 1
-enum/VREFINTRDYFR:
-  bit_size: 1
-  variants:
-    - name: NotReady
-      description: VREFINT is OFF
-      value: 0
-    - name: Ready
-      description: VREFINT is ready
-      value: 1
```
xoviat: Needs to be reversed because they're flipped.

PR #265, opened 2023-09-26
rcc: pipe through sel mux and generate ir

xoviat:


stm32-metapac-gen/src/lib.rs
```diff
@@ -318,6 +328,34 @@ impl Gen {
             let re = Regex::new("# *! *\\[.*\\]").unwrap();
             let data = re.replace_all(&data, "");
             file.write_all(data.as_bytes()).unwrap();
+
+            let mut data = String::new();
+
+            for (name, en) in ir.enums {
+                write!(
+                    &mut data,
+                    "const ENUM_{} = [",
```
Dirbaio, line 337: if you generate them like this, you're not going to be able to access them by name in `build.rs`.

I was thinking of dumping the entire chiptool IR here instead. so `build.rs` can do stuff like `for e in &regs.enums { for v in &e.variants { ... }}`

PR #267, opened 2023-09-27
add TAMP register block for g0, g4, l5, u5 and wl

oll3:
Adds TAMP register block for the devices where I could understand how it was mapped. Some devices seems to have their TAMP/backup/RTC more intertwined and not as a separate register block. Those are not included.

Also not really sure why only u5 and l5 was mapped automatically while I had to add them manually for g0, g4 and wl through extra/families. I guess some field is missing in one of the sources?

Mainly adding this for gaining access to backup registers. Tested on stm32wle5cc.

### DIFF
--- a/data/registers/tamp_g0.yaml
+++ b/data/registers/tamp_g0.yaml
@@ -1,0 +1,337 @@
+block/TAMP:
+  description: Tamper and backup registers
+  items:
+    - name: CR1
+      description: control register 1
+      byte_offset: 0
+      fieldset: CR1
+    - name: CR2
+      description: control register 2
+      byte_offset: 4
+      fieldset: CR2
+    - name: FLTCR
+      description: TAMP filter control register
+      byte_offset: 12
+      fieldset: FLTCR
+    - name: IER
+      description: TAMP interrupt enable register
+      byte_offset: 44
+      fieldset: IER
+    - name: SR
+      description: TAMP status register
+      byte_offset: 48
+      access: Read
+      fieldset: SR
+    - name: MISR
+      description: TAMP masked interrupt status register
+      byte_offset: 52
+      access: Read
+      fieldset: MISR
+    - name: SCR
+      description: TAMP status clear register
+      byte_offset: 60
+      access: Write
+      fieldset: SCR
+    - name: BKPR
+      description: TAMP backup register
+      array:
+        len: 5
+        stride: 4
+      byte_offset: 256
+      fieldset: BKPR
+    - name: HWCFGR2
+      description: TAMP hardware configuration register 2
+      byte_offset: 1004
+      access: Read
+      fieldset: HWCFGR2
+    - name: HWCFGR1
+      description: TAMP hardware configuration register 1
+      byte_offset: 1008
+      access: Read
+      fieldset: HWCFGR1
+    - name: VERR
+      description: EXTI IP Version register
+      byte_offset: 1012
+      access: Read
+      fieldset: VERR
+    - name: IPIDR
+      description: EXTI Identification register
+      byte_offset: 1016
+      access: Read
+      fieldset: IPIDR
+    - name: SIDR
+      description: EXTI Size ID register
+      byte_offset: 1020
+      access: Read
+      fieldset: SIDR
+fieldset/BKPR:
+  description: TAMP backup register
+  fields:
+    - name: BKP
+      description: BKP
+      bit_offset: 0
+      bit_size: 32
+fieldset/CR1:
+  description: control register 1
+  fields:
+    - name: TAMP1E
+      description: TAMP1E
+      bit_offset: 0
+      bit_size: 1
+    - name: TAMP2E
+      description: TAMP2E
+      bit_offset: 1
+      bit_size: 1
+    - name: ITAMP1E
+      description: ITAMP1E
+      bit_offset: 16
+      bit_size: 1
+    - name: ITAMP3E
+      description: ITAMP3E
+      bit_offset: 18
+      bit_size: 1
+    - name: ITAMP4E
+      description: ITAMP4E
+      bit_offset: 19
+      bit_size: 1
+    - name: ITAMP5E
+      description: ITAMP5E
+      bit_offset: 20
+      bit_size: 1
+    - name: ITAMP6E
+      description: ITAMP6E
+      bit_offset: 21
+      bit_size: 1
+fieldset/CR2:
+  description: control register 2
+  fields:
+    - name: TAMP1NOER
+      description: TAMP1NOER
+      bit_offset: 0
+      bit_size: 1
+    - name: TAMP2NOER
+      description: TAMP2NOER
+      bit_offset: 1
+      bit_size: 1
+    - name: TAMP1MSK
+      description: TAMP1MSK
+      bit_offset: 16
+      bit_size: 1
+    - name: TAMP2MSK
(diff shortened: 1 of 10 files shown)


xoviat:
Have you looked in the source files to see why chips.rs is not picking it up?


https://embassy.dev/book/dev/developer_stm32.html

oll3:
Seems `InstanceName="TAMP"` is not present in many of the xmls where the corresponding `TAMP` block is present in the svd.
Eg `stm32wle5.svd` has a `TAMP` peripheral but no  `InstanceName="TAMP"`  in `STM32WLE5CCUx.xml`.

How should this be handled?

data/registers/tamp_g0.yaml
```diff
@@ -0,0 +1,337 @@
...
+      bit_offset: 19
+      bit_size: 1
+    - name: ITAMP5E
+      description: ITAMP5E
+      bit_offset: 20
+      bit_size: 1
+    - name: ITAMP6E
+      description: ITAMP6E
+      bit_offset: 21
+      bit_size: 1
```
xoviat, line 104: consolidate to an array

xoviat, line 104: consolidate to an array

data/registers/tamp_g0.yaml
```diff
@@ -0,0 +1,337 @@
...
+      bit_offset: 1
+      bit_size: 1
+    - name: TAMP1MSK
+      description: TAMP1MSK
+      bit_offset: 16
+      bit_size: 1
+    - name: TAMP2MSK
+      description: TAMP2MSK
+      bit_offset: 17
+      bit_size: 1
```
xoviat, line 123: consolidate to an array

data/registers/tamp_g0.yaml
```diff
@@ -0,0 +1,337 @@
...
+  description: control register 2
+  fields:
+    - name: TAMP1NOER
+      description: TAMP1NOER
+      bit_offset: 0
+      bit_size: 1
+    - name: TAMP2NOER
+      description: TAMP2NOER
+      bit_offset: 1
+      bit_size: 1
```
xoviat, line 115: consolidate to an array

data/registers/tamp_g0.yaml
```diff
@@ -0,0 +1,337 @@
...
+      bit_offset: 17
+      bit_size: 1
+    - name: TAMP1TRG
+      description: TAMP1TRG
+      bit_offset: 24
+      bit_size: 1
+    - name: TAMP2TRG
+      description: TAMP2TRG
+      bit_offset: 25
+      bit_size: 1
```
xoviat, line 131: consolidate to an array

data/registers/tamp_g0.yaml
```diff
@@ -0,0 +1,337 @@
...
+      bit_offset: 19
+      bit_size: 1
+    - name: ITAMP5IE
+      description: ITAMP5IE
+      bit_offset: 20
+      bit_size: 1
+    - name: ITAMP6IE
+      description: ITAMP6IE
+      bit_offset: 21
+      bit_size: 1
```
xoviat, line 211: consolidate to an array

data/registers/tamp_g0.yaml
```diff
@@ -0,0 +1,337 @@
...
+      bit_offset: 19
+      bit_size: 1
+    - name: ITAMP5MF
+      description: ITAMP5MF
+      bit_offset: 20
+      bit_size: 1
+    - name: ITAMP6MF
+      description: ITAMP6MF
+      bit_offset: 21
+      bit_size: 1
```
xoviat, line 249: consolidate to an array

data/registers/tamp_g0.yaml
```diff
@@ -0,0 +1,337 @@
...
+      bit_offset: 20
+      bit_size: 1
+    - name: CITAMP6F
+      description: CITAMP6F
+      bit_offset: 21
+      bit_size: 1
+    - name: CITAMP7F
+      description: CITAMP7F
+      bit_offset: 22
+      bit_size: 1
```
xoviat, line 284: consolidate to an array

data/registers/tamp_g0.yaml
```diff
@@ -0,0 +1,337 @@
...
+      bit_offset: 20
+      bit_size: 1
+    - name: ITAMP6F
+      description: ITAMP6F
+      bit_offset: 21
+      bit_size: 1
+    - name: ITAMP7F
+      description: ITAMP7F
+      bit_offset: 22
+      bit_size: 1
```
xoviat, line 326: consolidate to an array

xoviat (changes requested):
any fields, etc that have a number suffix need to be consolidated to an array. see the hrtim_v1.yaml for examples of this.

xoviat:
> How should this be handled?

If it's not in the XML, you've done the right thing. I will merge after you arrayize the fields, per the review.

xoviat:
> I will merge after you arrayize the fields, per the review.

Also, to be clear I mean all of the files. Not just the one I reviewed.

oll3:
> any fields, etc that have a number suffix need to be consolidated to an array. see the hrtim_v1.yaml for examples of this.

I see, though many of these fields have flags which are not fully sequential. Should they be merged to arrays anyway?

Eg. ITAMPxIE from `tamp_wl.yaml` has flags on bits 18, 20, 21, 23:
```yaml
    - name: ITAMP3IE
      description: Internal tamper 3 interrupt enable
      bit_offset: 18
      bit_size: 1
    - name: ITAMP5IE
      description: Internal tamper 5 interrupt enable
      bit_offset: 20
      bit_size: 1
    - name: ITAMP6IE
      description: Internal tamper 6 interrupt enable
      bit_offset: 21
      bit_size: 1
    - name: ITAMP8IE
      description: Internal tamper 8 interrupt enable
      bit_offset: 23
      bit_size: 1
```

As an array I guess it would look like this, ignoring that offset 19 and 22 are not valid:
```yaml
    - name: ITAMPIE
      description: Internal tamper 3 interrupt enable
      bit_offset: 18
      bit_size: 1
      array:
        len: 6
        stride: 1
```

xoviat:
Use offsets in that case. Again, see hrtim for an example.

Dirbaio:
In this case the bit offsets line up nicely with the numbers, it's just that there's "holes" because some bits are unimplemented. In this case, just use a normal array. it's OK if you can write to the unimplemented bits, these are typically ignored anyway.

only use offsets if the bits really don't line up. For example if there was this:
- ITAMP1IE on bit 8
- ITAMP2IE on bit 9
- ITAMP3IE on bit 16
- ITAMP4IE on bit 17

oll3:
> In this case the bit offsets line up nicely with the numbers, it's just that there's "holes" because some bits are unimplemented. In this case, just use a normal array. it's OK if you can write to the unimplemented bits, these are typically ignored anyway.
> 
> (using offsets would prevent writing to unimplemented bits, but then it no longer holds that TAMPn is at index n-1, which is confusing)
> ...

Makes sense!

I have made repeated fields into arrays and also removed some more 1 bit enums which seemed fairly obvious. And renamed some enums with confusing names (mainly for u5).

Thank you both for taking the time to review and the guidance, really appreciate it! :star2:

xoviat (approved):

xoviat:
bender run

PR #279, opened 2023-10-12
Add OCTOSPI and OCTOSPIM peripherals.

noppej:
My objective here was to add OCTOSPIM (manager) and OCTOSPI (instances) as peripherals to STM32U5xx chips.

As suggested in the README, I also added where appropriate for other STM32 families.

This is my first attempt at a PR on this repo, and I am not 100% confident that I understood all the nuances correctly. Please feel free to comment liberally. I'd like to master this, because my ultimate goal is to also add the peripherals for GFXMMU,  GPU2D, TSC, and DSI in future PR's.


stm32-data-gen/src/chips.rs
```diff
@@ -468,6 +468,36 @@ impl PeriMatcher {
...
             ("STM32WL.*:TAMP:.*", ("tamp", "wl", "TAMP")),
+            (
+                "STM32L4R[59].*:OCTOSPIM:OCTOSPIM:octospi_v1_0.*",
+                ("octospim", "v1", "OCTOSPIM"),
+            ),
+            (
+                "STM32U5.*:OCTOSPIM:OCTOSPIM:octospi1_v3_0.*",
+                ("octospim", "v2", "OCTOSPIM"),
+            ),
+            (".*:OCTOSPIM:.*", ("octospim", "v2", "OCTOSPIM")), //Next chips releasing OCTOSPIM will default to v2.
```
Dirbaio, line 479: please don't add blanket matches like this. We don't know what next chips will do, it's better to not match registers and let the first person that needs to use octospi to check whether the regsiters are actually unchanged and add a new match rule.

noppej: Makes perfect sense. I think I misinterpreted other blanket matches (which defaults to v1, e.g. `EXTI`, but in hindsight, I can see they server a different purpose.

Dirbaio:
can you reformat the yamls with `chiptool fmt`? (sorry the CI doesn't catch this yet)

Dirbaio:
Seems there's something wrong with octospi_v3.yaml, many registers are "shuffled". For example:

- AR should have the "ADDRESS" field, but it has "DATA"
- ABR should have the "ALTERNATE" field, but it has "TIMEOUT"
- CCR should have "IMODE, IDTR, ..." but has "DCYC, DHQC, ..."

(pro tip, diff v1, v2, v3, v4 yamls looking for suspicious mistakes, a lot of these are actually SVD mistakes, like in this case!)

Dirbaio:
Given this, I think v3,v4 are actually the same, they could be merged. (v1->v2 does add many fields, v2->v3 removes DCR3.MAXTRAN)

could you
- Remove enums where 0=disabled, 1=enabled? like these

```yaml
enum/SMIE:
  bit_size: 1
  variants:
  - name: B_0x0
    description: Interrupt disabled
    value: 0
  - name: B_0x1
    description: Interrupt enabled
    value: 1

File: data/registers/octospi_v1.yaml [hunk 2]
```diff
@@ -0,0 +1,699 @@
...
+    fieldset: WIR
+  - name: WABR
+    description: write alternate bytes register
+    byte_offset: 416
+    fieldset: WABR
+  - name: HLCR
+    description: OCTOSPI HyperBus latency configuration register
+    byte_offset: 512
+    fieldset: HLCR
+  - name: HWCFGR
```
Dirbaio, line 94: these regs at the end of the 1kb peripheral register space are some kind of thing ST started adding around H7 to make peripherals "self describing". They're not specific to this peripheral, and we don't use them anyway. could you remove them?

data/registers/octospi_v1.yaml
```diff
@@ -0,0 +1,699 @@
...
+    bit_size: 1
+  - name: DMAEN
+    description: 'DMA enable In Indirect mode, the DMA can be used to input or output data via DR. DMA transfers are initiated when FTF is set. Note: Resetting the DMAEN bit while a DMA transfer is ongoing, breaks the handshake with the DMA. Do not write. this bit during DMA operation.'
+    bit_offset: 2
+    bit_size: 1
+  - name: TCEN
+    description: Timeout counter enable. This bit is valid only when the Memory-mapped mode (FMODE[1:0] = 11) is selected. This bit enables the timeout counter.
+    bit_offset: 3
+    bit_size: 1
+  - name: DQM
```
Dirbaio, line 209: this is an SVD mistake, the L4+ RM also calls this DMM, in-line with the other versions

data/registers/octospi_v1.yaml
```diff
@@ -0,0 +1,699 @@
...
+    bit_size: 1
+  - name: FRCK
+    description: Free running clock. This bit configures the free running clock.
+    bit_offset: 1
+    bit_size: 1
+  - name: CSHT
+    description: Chip-select high time CSHT + 1 defines the minimum number of CLK cycles where the chip-select (NCS) must remain high between commands issued to the external device. ...
+    bit_offset: 8
+    bit_size: 3
+    enum: CycleShift
```
Dirbaio, line 270: this enum is for another field

data/registers/octospi_v1.yaml
noppej: I thought I caught that, but it slipped through.

Dirbaio:
also octospim v1 is missing CR, according to the L4 RM it is there. So v1/v2 are identical, there's just one octospim version.

noppej:
> also octospim v1 is missing CR, according to the L4 RM it is there. So v1/v2 are identical, there's just one octospim version.

Cool. I believe I'm beginning to see the pattern here ... consolidate yamls and review diffs against RM, repeat until diffs are absolute minimal, then repeat. At first there were so many diffs I couldn't see the forest for the trees, but it is falling in place now.

Dirbaio:
yeah, exactly. You can't trust SVDs at all, they're really really bad. Always doublecheck all differences with the RMs :smiling_face_with_tear:

noppej:
> yeah, exactly. You can't trust SVDs at all, they're really really bad. Always doublecheck all differences with the RMs 🥲

OK. So I have one more diff that was wrong in the svd.
`DCR1/CSHT`: In L4 it was bit size 3, but RM is 6, so now it matches with others.

Dirbaio:
you got it the other way around, v1 has MAXTRAN and v2 removes it.

Dirbaio (approved):
thank you! :partying_face:

PR #290, opened 2023-10-16
Add TSC (touch screen controller) peripheral

noppej:
Note: Some field sets have potential to be collapsed to two dimensional arrays, but `chiptool` doesn't support this.


Dirbaio:
thank you! :D

PR #297, opened 2023-10-20
rcc: solve hashmap determinism for good

xoviat:


Dirbaio:
this'll pick the smallest key consistently, making it deterministic, yes. 

However, in the case of RCC enable/reset bits, it is a bug if there are two registers matching one peripheral. Picking the smallest key is not necessarily correct, it just hides the problem. I think we should error out in that case, and then fix all the instances of the error.

PR #300, opened 2023-11-01
Add enum/ADCSET to rcc_l5.yaml

ShakenCodes:
enum/ADCSET was omitted on the STM32L5 processor family.

Change per discussion with dirbaio on Matrix channel

### DIFF
--- a/data/registers/rcc_l5.yaml
+++ b/data/registers/rcc_l5.yaml
@@ -1931,7 +1931,7 @@ enum/ADCSEL:
   - name: DISABLE
     description: No clock selected
     value: 0
-  - name: PLL1_Q
+  - name: PLLSAI1_R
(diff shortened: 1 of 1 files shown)


data/registers/rcc_l5.yaml
```diff
@@ -1924,6 +1925,18 @@ fieldset/SECSR:
     description: RMVFSECF
     bit_offset: 12
     bit_size: 1
+enum/ADCSEL:
+  bit_size: 2
+  variants:
+  - name: DISABLE
+    description: No clock selected
+    value: 0
+  - name: PLL1_Q
```
Dirbaio, line 1934: should be PLLSAI1_R according to the RM.

Dirbaio (approved):

PR #302, opened 2023-11-05
add "enum debug" helper macro

xoviat:
the intention is to expand this to become an independent crate, eventually


Dirbaio (approved):
lgtm, nice idea! makes dumping much cleaner

PR #306, opened 2023-11-06
Rework DAC support for all STM32s

adamgreig:
It's very similar to v3, but has a few fields added for some new functionality, and the TSEL enum is now identical between channel 1/2 (but very slightly different DAC1/2/3/4  for variant 1 and 15 - details in description).

All fields that existed in v3 are identical in v4, so the diff to `dac_v3.yaml` is fairly clean, ~~except for v3's TSEL1/2 being merged to a TSEL array in v4~~ for compatibility with embassy-stm32, I split them back into TSEL1/2 and made identical TSEL1/2 enums. It's not great, though - should these be combined and I just teach embassy-stm32 how to use a single TSEL field for v4?

It seems only G4 uses dacv4, and the newer H5 uses dacv5.


adamgreig:
I've gone through every RM and checked the DACs and updated with what I think is a decent new set of DAC versions, v1-v7. There's not much intentional difference between the YAMLs, and zero unintentional differences, so it should be easy enough to maintain with any luck.

This is the summary:

image

I've removed the TSEL enum so the field can be an array, and will update the HAL to implement it there instead. What we had currently was wrong on a lot of chips because it changes within the same IP version:

image

The only difference from the RMs I'm aware of is that the U5 has a new "AUTOMODE" register which is not implemented here; it would need to be a new version (perhaps v7 and the g4 bumps to v8) as it's not in the H5 that's otherwise the same. I could add this too if it's desired.

Dirbaio:
bender run

Dirbaio (approved):
LGTM. (will wait for https://github.com/embassy-rs/embassy/pull/2199 before merging to not break embassy-stm32)

PR #308, opened 2023-11-15
fdcan: add H7 support; fix regs for others; add message RAM blocks

tcbennun:
This is an attempt to fix the FDCAN peripheral support, which is not generated in metapac for some chips, and for others, is generated but incorrect.

There's a related (abandoned?) draft PR by another author here: https://github.com/embassy-rs/embassy/pull/1692

So firstly, the Cube XMLs refer to "fdcan_v1_0" and "fdcan_v1_1" but these seem to have no bearing on the actual registers used. Thus chips.rs should make no distinction between v1_0 and v1_1. Before this PR, v1_1 is ignored, so some chips don't get FDCAN generated.

The current YAML seems to have been generated from a G4 SVD, but this series' SVDs consistently have several errata.

I have therefore replaced can_fdcan.yaml with can_fdcan_v1.yaml, built from an H5 SVD which appears to match the RMs of G0, G4, H5, L5 and U5 chips.

Meanwhile, the H7 series has a totally different FDCAN, so I've added a separate YAML for that.

I've added defs for the message RAM blocks, and code to generate them (they're not listed as periphs generally).

### DIFF
--- a/data/registers/rcc_g4.yaml
+++ b/data/registers/rcc_g4.yaml
@@ -1006,17 +1006,18 @@ fieldset/CCIPR:
     bit_offset: 18
     bit_size: 2
   - name: SAI1SEL
-    description: Low power timer 2 clock source selection
+    description: SAI1 clock source selection
     bit_offset: 20
     bit_size: 2
   - name: I2S23SEL
-    description: SAI1 clock source selection
+    description: I2S23 clock source selection
     bit_offset: 22
     bit_size: 2
   - name: FDCANSEL
-    description: SAI2 clock source selection
+    description: FDCAN clock source selection
     bit_offset: 24
     bit_size: 2
+    enum: FDCANSEL
   - name: CLK48SEL
     description: 48 MHz clock source selection
     bit_offset: 26
@@ -1351,6 +1352,18 @@ enum/ADCSEL:
     value: 1
   - name: SYS
     description: System clock selected as ADC clock
+    value: 2
+enum/FDCANSEL:
+  bit_size: 2
+  variants:
+  - name: HSE
+    description: HSE used as FDCAN clock source
+    value: 0
+  - name: PLL1_Q
+    description: PLLQCLK used as FDCAN clock source
+    value: 1
+  - name: PCLK1
+    description: PCLK used as FDCAN clock source
     value: 2
 enum/CLK48SEL:
   bit_size: 2
--- a/data/registers/rcc_l5.yaml
+++ b/data/registers/rcc_l5.yaml
@@ -1374,6 +1374,7 @@ fieldset/CCIPR:
     description: FDCAN clock source selection
     bit_offset: 24
     bit_size: 2
+    enum: FDCANSEL
   - name: CLK48SEL
     description: 48 MHz clock source selection
     bit_offset: 26
@@ -1937,6 +1938,18 @@ enum/ADCSEL:
   - name: SYS
     description: SYSCLK clock selected
     value: 3
+enum/FDCANSEL:
+  bit_size: 2
+  variants:
+  - name: HSE
+    description: HSE clock selected
+    value: 0
+  - name: PLL1_Q
+    description: PLL "Q" clock selected
+    value: 1
+  - name: PLLSAI1_P
+    description: PLLSAI "P" clock selected
+    value: 2
 enum/CLK48SEL:
   bit_size: 2
   variants:


tcbennun:
Note: The H7 chips have two additional sets of associated registers (TTCAN and CCU). I don't know how or whether to map these.

tcbennun:
I'd welcome a review if possible 🙂 

I have a couple things I'm unclear how to solve:

1. I cannot find message RAM offsets in any data source. I am unsure where to define them. There's a constant offset from the base address, per FDCAN peripheral. This currently lives in `chips.rs`:

https://github.com/embassy-rs/stm32-data/blob/27c71ac451ff9011fca1d09dc3b7c1d92929bfb7/stm32-data-gen/src/chips.rs#L965

2. for H7, the RX1FS reg has one extra field compared to RX0FS, which is just annoying and messes with arrayifying the regs. Currently this field is omitted:

https://github.com/embassy-rs/stm32-data/blob/27c71ac451ff9011fca1d09dc3b7c1d92929bfb7/data/registers/can_fdcan_h7.yaml#L947

---

Also in the main `embassy` repo I will need to figure out how to associate each message RAM with the correct FDCAN periph but I'm pretty sure that isn't relevant to `stm32-data`

kevswims:
@tcbennun for the calculation of the RAM offsets where did the 0x350 value come from? That seems to be wrong for the G4 at least where each message RAM is 1K

image

kevswims:
@tcbennun Do you have any embassy code that is working to test this with? I've got a G4 with CAN hardware that I could give this a try on.

tcbennun:
> @tcbennun for the calculation of the RAM offsets where did the 0x350 value come from? That seems to be wrong for the G4 at least where each message RAM is 1K

The memory layout table is misleading. In RM0440, see instead section 44.3.3, pp. 1952-3. Each FDCAN module allocates 212 words (848 = 0x350 bytes) in the message RAM.

The RM states that this size is used to calculate the offset for subsequent FDCANs. However, just to muddy the waters further, it then claims that the offset should be 0x354 instead of 0x350. Sigh.

We have further evidence for 0x350 being correct in the official HAL.

The 0x350 size is calculated in macros here:

https://github.com/STMicroelectronics/stm32g4xx_hal_driver/blob/7c2e19ec21714f0d993ae5bfe8859a45debcfdd6/Src/stm32g4xx_hal_fdcan.c#L235

Which is then used to calculate the offsets at the starts of this function:

https://github.com/STMicroelectronics/stm32g4xx_hal_driver/blob/7c2e19ec21714f0d993ae5bfe8859a45debcfdd6/Src/stm32g4xx_hal_fdcan.c#L3397C13-L3397C13

tcbennun:
> @tcbennun Do you have any embassy code that is working to test this with? I've got a G4 with CAN hardware that I could give this a try on.

I have an H5 with only one FDCAN which I've got working (at least on transmit). I should have time today to put that somewhere for you!

tcbennun:
> @tcbennun Do you have any embassy code that is working to test this with? I've got a G4 with CAN hardware that I could give this a try on.

@kevswims 

Clone this repo: https://github.com/tcbennun/embassy-fdcan-mvp then clone my `stm32-data` fork adjacent to it

It only does TX and it's setup for an H503 but you should be able to adapt it fairly easily

It uses my fork of embassy ( https://github.com/MaxiluxSystems/embassy/tree/feature/fdcan ) which has a WIP FDCAN driver

aaamorgan:
> Clone this repo: https://github.com/tcbennun/embassy-fdcan-mvp then clone my `stm32-data` fork adjacent to it
> 
> It only does TX and it's setup for an H503 but you should be able to adapt it fairly easily
> 
> It uses my fork of embassy ( https://github.com/MaxiluxSystems/embassy/tree/feature/fdcan ) which has a WIP FDCAN driver
> 
> Remember to generate the stm32 pac before building the test project!

@tcbennun I was able to get an example running on a G4 using your stm32 pac. I did have to make additional changes to your embassy fork to get it working - basically just carrying changes you made over to the g4 rcc. I also added an example that has external loopback working for FDCAN2 and FDCAN3 separately. All this is on my fork https://github.com/lucimobility/embassy/tree/feature/fdcan.

I also tested normal receiving and transmitting independently for both FDCANs using external CAN hardware, but I ran into issues whenever I tried using multiple FDCAN instances simultaneously. Even just initializing a second `FdCan` instance without using it at all seems to break the first instance's ability to read or write. Is this something you could test on your hardware to see if it's doing the same thing?

tcbennun:
> @tcbennun I was able to get an example running on a G4 using your stm32 pac. I did have to make additional changes to your embassy fork to get it working - basically just carrying changes you made over to the g4 rcc. I also added an example that has external loopback working for FDCAN2 and FDCAN3 separately. All this is on my fork https://github.com/lucimobility/embassy/tree/feature/fdcan

Thanks for testing! Much appreciated.

> I also tested normal receiving and transmitting independently for both FDCANs using external CAN hardware, but I ran into issues whenever I tried using multiple FDCAN instances simultaneously. Even just initializing a second `FdCan` instance without using it at all seems to break the first instance's ability to read or write. Is this something you could test on your hardware to see if it's doing the same thing?

Soon I'm gonna have a G0 board with two FDCANs, so I'll be looking to get that working! Hopefully whatever I find can translate to G4.

TheEppicJR:
> I also tested normal receiving and transmitting independently for both FDCANs using external CAN hardware, but I ran into issues whenever I tried using multiple FDCAN instances simultaneously. Even just initializing a second `FdCan` instance without using it at all seems to break the first instance's ability to read or write. Is this something you could test on your hardware to see if it's doing the same thing?
I just got to testing with both FDCAN's on the H7 and I see the same behavior

kevswims:
I did a bit of digging on the multiple CAN's not working and I think I know what is going on but I am not quite sure what the best behavior is to fix it.

The build.rs automatically implements the `rcc::sealed::RccPeripheral` trait which has a `enable_and_reset_with_cs` method. For some reason, ST combined the enable and reset registers for all of the FDCAN instances into a single register. As a result, when the second instance is initialized, it will reset all of the instances and clear the configuration of the first instance.

The part I am not sure about, is if we make that function not do the reset and enable if it is already enabled then it doesn't really do what it says it does since it would never reset the peripheral if it has been enabled even when using a single peripheral. To test this out, we can try making the peripheral that is initialized second no-op that function or something.

Dirbaio (approved):
:rocket:

PR #309, opened 2023-11-16
STM32L151C(6,8,B)T ADC

CaptainMaso:
Seems like this peripheral is only used for these very specific chips, but I've also done some work towards getting them working in Embassy too.


stm32-data-gen/src/rcc.rs
```diff
@@ -373,6 +373,11 @@ impl PeripheralToClock {
             return clocks.get("ADC");
         }
 
+        // Absolute fallback, match against the clocks for just the first ADC
```
Dirbaio, line 376: could you make this a bit more restrictive, like `peri_name == "ADC"`? Otherwise it's a bit risky, it could cause silent failures like e.g. `ADC3` to fall back to `ADC1` clock bit.

stm32-data-gen/src/chips.rs
```diff
@@ -932,9 +934,43 @@ fn process_core(
...
+        "SYSCFG",
+        "EXTI",
+        "FLASH",
+        "DBGMCU",
+        "CRS",
+        "PWR",
+        "AFIO",
+        "BKP",
+        "USBRAM",
+        "VREFINTCAL",
```
Dirbaio, line 973: This causes adding `VREFINTCAL` to many more chips, like F7s. is it intentional, have you checked it's correct?

CaptainMaso: The F7's weren't intentional, I'll check now if it's valid and filter it out if not.

CaptainMaso: These values are all defined in the sources/headers files for those specific chips (which was the intentional source for the L1 series), but I can't find a reference to it anywhere in the reference manuals.
I can try and strip it out if necessary but it seems fairly legitimate.

stm32-data-gen/src/rcc.rs
CaptainMaso: Done in the latest commit

Dirbaio (approved):
thank you!

PR #310, opened 2023-11-16
split adc_v2 to f2/f4 and f7 as they have incompatible EXTSEL/JEXTSEL definition

aurelj:
Currently, the EXTSEL enum is defined to match the F2/F4 series (incorrect on F7 series) and the JEXTSEL enum is defined to match the F7 series (incorrect on the F2/F4 series).

The worse is that those enums share some fields with the same name but different values.
I tried to use Jextsel::TIM5TRGO on a F4 and it compiled fine but it took me a long time to understand why it didn't work.
Issue was that Jextsel::TIM5TRGO was defined to 12 (correct on F7) while it should be define to 11 on F4 !


Dirbaio:
this is the same issue we've run into with DAC at #309 . The conclusion was that it's better to completely remove these enums from the PAC, These enums change A LOT between chips, if we keep them in the PAC we'll end up with too many versions that are identical except these enums.

The embassy-stm32 HAL can then use `cfg`s to add the right enums later. (though embassy-stm32 doesn't support adc triggers for now, so no change is required for now)

could you change it to do this instead? ideally on all ADC versions for consistency.

aurelj:
Well,  these enums are at least consistent in the whole F4 family and the whole F7 family (but they differ between families).

I agree that it is not nice to have to duplicate those files just for one or two enums. Some kind of mecanism to "include" a YAML file from another one would be nice to avoid this kind of duplication.

Anyway, if you think it is better to just remove those enums, it is also fine with me, so I opened PR #312 to remove them.

aurelj:
Closed in favor of #312

Issue #293, opened 2023-10-18
STM32G474CB Missing Quad SPI peripheral


Tupelov:
For whatever reason, the generater only outputs the qspi interrupt but not the actual peripheral itself.

xoviat:
looks like we don't have an implementation for that version at all, so it will need to be added

Tupelov:
> looks like we don't have an implementation for that version at all, so it will need to be added

I've tried various times to create a v4 version of the peripheral but for whatever reason it won't show up in the output. From what I could tell, only the bit width of a single register was changed and to my knowledge they changed the name of the qspi enable register but even after modifying the RCC yaml, I couldn't get it to work.
Could there be something else besides the chips.rs and quadspi_v4.yaml that I need to edit?

dlaw:
I just used quadspi via stm32-metapac on STM32G473, so I can confirm that #311 fixed this issue and this issue can be closed.

PR #311, opened 2023-11-29
Fixed generation for quadspi v4 for STM32g474 family

Tupelov:
The main cause of issue #293 was that in addition to a new version of the QUADSPI peripheral, it had a different name inside the C header that defined its base address, leading to it never being generated.

### DIFF
--- a/data/registers/quadspi_v4.yaml
+++ b/data/registers/quadspi_v4.yaml
@@ -1,0 +1,291 @@
+block/QUADSPI:
+  description: QuadSPI interface
+  items:
+  - name: CR
+    description: control register
+    byte_offset: 0
+    fieldset: CR
+  - name: DCR
+    description: device configuration register
+    byte_offset: 4
+    fieldset: DCR
+  - name: SR
+    description: status register
+    byte_offset: 8
+    access: Read
+    fieldset: SR
+  - name: FCR
+    description: flag clear register
+    byte_offset: 12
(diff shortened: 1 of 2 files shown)


Tupelov:
Im not sure how to get the checks to work

Dirbaio:
rebase on top of latest main, then force push

Tupelov:
Hold on, I just realized there is both a `QSPI_BASE` and `QSPI_R_BASE`, I think that my code may be incorrect.

Tupelov:
image
I noticed that fmc also has a similar setup, however I tried doing what was done with FMC and I got the same address which is `0x90...` In the svd the peripheral address is `0xA0001000`  so I think I will instead pass QSPI_R to the header function instead. Do you think that  would be the correct solution in this situation?

Tupelov:
Looking over it, `0x90..` seems to be the base address for the memory mapped mode

Tupelov:
image

Tupelov:
Since it appears that the address is fixed for each chip, I think this should be implemented in embassy-stm32, so I'll swap it to QSPI_R

Dirbaio:
Thanks for the PR!

This issue actually affects more chips than g474, it's nicer to fix it at once for all chips. Also, the registers for QSPI in G4 are actually the same as other chips, so a new version wasn't needed. I've opened #313 with the fixes.

PR #320, opened 2023-12-24
spi-cleanup

eZioPan:


Dirbaio:
most enums this is removing are not "0=disabled, 1=enabled", we should not remove them.

For example 0=LSBFIRST, 1=MSBFIRST is a bit that chooses between two behaviors, it's clearer with an enum. It's not 0 = don't enable something, 1 = enable something.

eZioPan:
understand, I will try restore some enum.

Dirbaio (approved):
thanks!

PR #321, opened 2023-12-24
refactor with clippy

eZioPan:
not some serious change, just followed `cargo clippy`'s advices.
But still need some careful check.


eZioPan:
CI seems good, don't known if CI use my PR to build the binary tools.

PR #322, opened 2023-12-24
refactor with clippy

eZioPan:
not some serious change, just followed `cargo clippy`'s advices.
But still need some careful check.


Dirbaio:
CI shows no diff in the output, so it looks good!

PR #325, opened 2023-12-26
Add rtc registers for stm32l5

chrenderle:
I added the rtc registers for the stm32l5 series. As far as I understand there are no generated data in this repo anymore? At least not for this change I think. So there is no separate commit to put the generated data in as described in the README.

I also made some minor changes to the rtc_v3u5 registers. Mainly I used the array feature for all ALARM A and B registers. This was not used for all registers, but only for some.


Dirbaio (approved):
thank you!

PR #327, opened 2023-12-29
ci: use stable 1.75

eZioPan:


eZioPan:
no, there are still nightly features

PR #329, opened 2024-01-01
syscfg-cleanup

eZioPan:
remove enum FPU_IE[0-5], PAGE[0-15]_WP and VBAT_MON


eZioPan:
Should merge fields into array… will need to find correct way to write down a "transform"

Dirbaio (approved):

PR #330, opened 2024-01-01
Eth v1 cleanup

eZioPan:
Almost all modification (other than description fix) is done by **ETH_v1.yaml** transform file.


data/registers/eth_v1a.yaml
```diff
@@ -210,6 +186,20 @@ block/ETHERNET_MAC:
     byte_offset: 452
     access: Read
     fieldset: MMCRGUFCR
+  - name: MACAHR
```
Dirbaio, line 189: should also include `MACA0*`

Dirbaio: ah no, I see they're different

Dirbaio (approved):
thank you!

PR #344, opened 2024-01-04
reconstruct TIMER.yaml transform file

eZioPan:


data/registers/timer_v1.yaml
```diff
@@ -755,15 +753,6 @@ enum/ETF:
   - name: FDTS_Div32_N8
     description: fSAMPLING=fDTS/32, N=8
     value: 15
-enum/ETP:
```
Dirbaio, line 758: i'd keep this one, "polarity = false" is not very clear. If the bit was named "invert" then removing it would make more sense.

eZioPan:
oh, no, I shouldn't remove both ETP and MSM, both field name don't fit into true/false values.

Reconstruct last modification into a transform yaml should be enough.

Dirbaio (approved):

PR #350, opened 2024-01-04
add cordic_v1 for stm32u5

tyler-gilbert:
This MR adds support for the cordic_v1 unit for the stm32u5.

This is the last MR I have for now.

### DIFF
--- a/data/registers/cordic_v1.yaml
+++ b/data/registers/cordic_v1.yaml
@@ -1,0 +1,209 @@
+block/CORDIC:
+  description: CORDIC co-processor.
+  items:
+  - name: CSR
+    description: Control and status register.
+    byte_offset: 0
+    fieldset: CSR
+  - name: WDATA
+    description: Argument register.
+    byte_offset: 4
+    fieldset: WDATA
+  - name: RDATA
+    description: Result register.
+    byte_offset: 8
+    fieldset: RDATA
+fieldset/CSR:
+  description: Control and status register.
+  fields:
+  - name: FUNC
+    description: Function.
+    bit_offset: 0
+    bit_size: 4
+    enum: FUNC
+  - name: PRECISION
+    description: Precision required (number of iterations/cycles), where PRECISION = (number of iterations/4).
+    bit_offset: 4
+    bit_size: 4
+    enum: PRECISION
+  - name: SCALE
+    description: Scaling factor (2^-n for arguments, 2^n for results).
+    bit_offset: 8
+    bit_size: 3
+  - name: IEN
+    description: Enable interrupt.
+    bit_offset: 16
+    bit_size: 1
+  - name: DMAREN
+    description: Enable DMA wread channel.
+    bit_offset: 17
+    bit_size: 1
+  - name: DMAWEN
+    description: Enable DMA write channel.
+    bit_offset: 18
+    bit_size: 1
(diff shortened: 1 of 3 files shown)


lights0123:
Are you working on a higher-level CORDIC driver? I was going to take a shot at it if you aren't.

stm32-data-gen/src/chips.rs
```diff
@@ -514,6 +514,7 @@ impl PeriMatcher {
             ("STM32L4.*:GFXMMU:.*", ("gfxmmu", "v1", "GFXMMU")),
             ("STM32U5.*:GFXMMU:.*", ("gfxmmu", "v2", "GFXMMU")),
             ("STM32U5.*:ICACHE:.*", ("icache", "v1", "ICACHE")),
+            ("STM32U5.*:CORDIC:.*", ("cordic", "v1", "CORDIC")),
```
lights0123, line 517:
```suggestion
            (".*:CORDIC:.*", ("cordic", "v1", "CORDIC")),
```

All STM chips with CORDIC support (G4, H7, H5, U5) have the same registers, so let's just expose it for all of them.

data/registers/cordic_v1.yaml
```diff
@@ -0,0 +1,209 @@
...
+  description: CORDIC co-processor.
+  items:
+  - name: CSR
+    description: Control and status register.
+    byte_offset: 0
+    fieldset: CSR
+  - name: WDATA
+    description: Argument register.
+    byte_offset: 4
+    fieldset: WDATA
```
lights0123, line 11:
```suggestion
```

data/registers/cordic_v1.yaml
```diff
@@ -0,0 +1,209 @@
...
+  - name: RDATA
+    description: Result register.
+    byte_offset: 8
+    fieldset: RDATA
```
lights0123, line 15:
```suggestion
```

data/registers/cordic_v1.yaml
```diff
@@ -0,0 +1,209 @@
...
+    description: Function result.
+    bit_offset: 0
+    bit_size: 32
+fieldset/WDATA:
+  description: Argument register.
+  fields:
+  - name: ARG
+    description: Function input arguments.
+    bit_offset: 0
+    bit_size: 32
```
lights0123, line 83:
```suggestion
```

lights0123 (changes requested):
These make it rather inconvenient to use as newtypes are generated for simple word types. By removing these, it makes it work like the CRC peripheral so that those registers are just exposed as integers:
image

data/registers/cordic_v1.yaml
tyler-gilbert: Will fix in the next commit.

data/registers/cordic_v1.yaml
tyler-gilbert: Got it.

tyler-gilbert:
@lights0123, I will commit the changes you suggested shortly. I haven't worked on a higher level driver and don't plan to at the moment, but from what I read it should be very easy to do that.

tyler-gilbert:
@lights0123, Updated this MR with your suggestions.

Dirbaio (approved):
thank you!

PR #357, opened 2024-01-07
apply chiptool validation

eZioPan:
~~add bit_size check:~~

~~1. max value of variant shouldn't overflow enum bit_size~~
~~2. field and enum should have same bit_size~~

Update after [comment](https://github.com/embassy-rs/stm32-data/pull/357#issuecomment-1880164908):

apply chiptool::validate::validate when parse yaml, but disable register overlap check and field overlap check

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,10 +12,58 @@ dependencies = [
 ]
 
 [[package]]
-name = "anyhow"
-version = "1.0.71"
+name = "anstream"
+version = "0.6.5"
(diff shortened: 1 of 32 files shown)


Dirbaio:
thanks for fixing all these!

Code for checking for these mistakes already exists in chiptool, see https://github.com/embassy-rs/chiptool/blob/main/src/validate.rs . you can do `chiptool check data/registers/*.yaml`, it's also available in chiptool as a lib. Could you change it to use chiptool, so we don't duplicate the code?

PR #360, opened 2024-01-12
rework on entire TIM mapping based on AN4013

eZioPan:
I find there are some issues on embassy-rs/embassy that related to missing/error mapping of some TIM. So I try to rework entire TIM mapping based on Rev.10 of [AN4013](https://www.st.com/resource/en/application_note/an4013-stm32-crossseries-timer-overview-stmicroelectronics.pdf).

This PR is based on AN4013 and with eye check on CI json diff. It should add some missing LPTIM mapping, change some TIM_GP16 to TIM_ADV, and exchange some TIM_GP32 and TIM_GP16.

If my assumption

* "1-channel" and "2-channel" in AN4013 is "TIM_GP16"
* and "1-channel with complementary output" and "2-channel with complementary output" in AN4013 is "TIM_ADV"

is correct. Then this PR *should* work.

I didn't check every RM, *hope this won't fry someone's chips. XD*

PR #363, opened 2024-01-19
timer v2

eZioPan:
`DCR` and `DMAR` registers has different address for F4 and G4 serial. And G4 reuse `DCR` and `DMAR` registers address for new `CCR5` and `CCR6` registers. Thus it's a new version.

`timer_v2.yaml` doesn't link to any actual MCU, will wait until

- #360 merged, and check which serial it belongs to.


eZioPan:
close for now, re-open when ready

PR #364, opened 2024-01-20
timer v2

eZioPan:
Create tim_v2 based on [AN4013 Rev.10] and [RM0481 H5], should cover most newer MCUs' timers.

Try best effort to make registers access unified.
~No plan to merge them into one file, too fragmentary to do so.~ (try make one)

~Haven't link to any actual MCUs. Don't know if I should port it to timer_v1.~
Remake timer_v1 from timer_v2, based on [RM0399 H7].
Remake entire TIM/LPTIM/HRTIM mapping based on [AN4013 Rev.10].

------

The TIM periperals are more complicated than I though.
Here is a diagram I drew for the block/fieldset inherit. ~Red/Green/Blue path is the timer yaml actually been used, grey path is also usable but not used in yaml.~
If someone want to modify it, just download the svg and load it into draw.io.

timer drawio

------
------

~As for the HAL implement, I would like to leave it to someone more familiar to `embassy-stm32` crate. My ability is not capable enough to do this modification.~

try create one... it much harder task than clean PAC...

[AN4013 Rev.10]: https://www.st.com/resource/en/application_note/an4013-stm32-crossseries-timer-overview-stmicroelectronics.pdf

[RM0481 H5]: https://www.st.com/resource/en/reference_manual/rm0481-stm32h563h573-and-stm32h562-armbased-32bit-mcus-stmicroelectronics.pdf

[RM0399 H7]: https://www.st.com/resource/en/reference_manual/rm0399-stm32h745755-and-stm32h747757-advanced-armbased-32bit-mcus-stmicroelectronics.pdf


eZioPan:
got to add some virtual compatibility layer to make HAL implement easier and safer

eZioPan:
actual inheritance
timer drawio

eZioPan:
Oh no, TIM_CH2 of STM32L0 is special, it has Exteral Trigger Pin ...

Dirbaio (approved):
This is incredibly thorough, it's evident you've put a lot of work into it. Thanks for this! 

(why are stm32 timers so complicated? :sob:)

eZioPan:
@Dirbaio Finally we made it! CONGRATS!

> (why are stm32 timers so complicated? 😭)

I thought it would be as simple as changing a few lines of `chip.rs`, to simply remap some TIM. Then I find STM32H5 have move `DCR` and `DMAR` registers to very end of the ADV TIM periperial, and reuse original address as other reigsters, so I have to split a new version of timer.

When I make new timer version, I though "I have already make a new version, why not make it more precise?". From that moment, I start to make 8 diffrent TIM peripherals.

And after that, I ask you in matrix about "should I merge them into one peripheral?", and got your positive answer. Then I start to merge them together.

It actually a little bit hard for me at beginning, since I hadn't though about the relationship among all these TIMs. But timer traits in embassy-stm32 gave me some hints, so I followed the traits hierarchy, and merged them together. And I'm actually benifit from mergering it, since mergering them means do another round of check with another perspective. And I did find some bug inside it.

After I merged them together for the first time, I was suggling to make a decision whether or not make a upgrade to embassy-stm32. Since I was a little too tired on making TIM PAC, but also my PR is a breaking change, I guess no one would love to jump into this mess and make a HAL change. So I made my mind to do a PR on embassy-stm32.

I have to say, embassy-stm32 is ingenuity designed when using it, but a great maze when modify it. I can't remember how many times I lost inside "sealed traits" / "macros" / "macros inside macros" / "macros implment traits" / "macros and traits from/to other files (even build.rs)". And I can't remember how many times I want to give it up. But after all I find a correct way to make it.

And for the HAL implemention, I splitted `BASIC_NO_CR2` from `BASIC` for DMA trait (1CH and 2CH don't support DMA). And do some "TIM_A should re-based-on TIM_B, rather than TIM_C" thing, after I think about `pin_trait!()`.

No matter what, we are arriving at here. CONGRATS AGAIN!

PR #366, opened 2024-01-23
Update U5 Octospi mapping

kkoppul2:
Update the PERIMAP to correctly associate the U535 and U545 chips with the appropriate version of the octospi implementation. These two sub-families use a different implementation than the rest of the U5 set of chips leading from the fact that only a single octospi peripheral is available on these chips, rather than two with the others.


Dirbaio (approved):
thanks!

PR #368, opened 2024-01-27
Adding SAI v1.2 peripheral support

### LINKED ISSUE #367: Missing implementation of SAI for sai1_v1_2
I'm working with an [STM32F723 which supports SAI, but the version is v1_2.](https://github.com/embassy-rs/stm32-data-sources/blob/796a8322b8f6e4329002a1fb49438aec95dae812/cubedb/mcu/STM32F723I(C-E)Kx.xml#L38-L39) If I'm following the logic of how the stm32-metapac is created, there should be a matching entry in the `PERIMAP`, as seen here:
https://github.com/embassy-rs/stm32-data/blob/6d4ee04c7d5fd26638c520a0377614dc25339abb/stm32-data-gen/src/chips.rs#L275-L279

I have a few questions around the implementation here. First is the SAI versions (V1 through V4). Are these completely unrelated to the STM implementations? The mapping of `sai1_v1_0` -> ` v1` and `sai1_v1_1` -> `v2` but `sai1_v2_0` -> `v1`? From the semver standpoint, none of this makes sense. Second, the cubedb file has `sai1_v1_2` for my target MCU, does anyone know if STM publishes the changelist for what v1_2 would include? If it's actually just a minor revision, then I'd be happy with adding a mapping to v2(?) to get the basic functionality I need. If it's not just a minor revision, then I suppose I'd have to add a V5 implementation here?

Cryowatt:
I started with generating what I had expected to be SAI v5 from the `F7x3` series, but after updating the SAI transformation to catch all of the enable/disable flags that were seemingly hand-edited out of the other SAI implementations, I realized that post-cleanup V5 was identical to V2. So I've added a new entry to PERIMAP. And I've included the modified SAI transformation configuration.

This commit closes #367


Dirbaio:
thank you!

PR #370, opened 2024-01-27
Fix register address mapping

kkoppul2:
Octospi peripheral register address is getting set to the memory-mapped region for the U5 family. This alternate mapping for the peripheral corrects the register address that is generated.


kkoppul2:
Validated updated addresses against the reference manual for each family of chips

Dirbaio:
thanks!

PR #372, opened 2024-01-27
Fix crc version mapping

TaylorCarpenter:
CRC version mapping is now done by family, rather than the IP version listed in MCU XML, because that was incorrect (not to mention the fact that the "IP versions" were integration tests?)
Versions now match what is on spreadsheet


TaylorCarpenter:
There are some chips that weren't explicitly called out in the spreadsheet, so we can fall back on the original checks against 'integtest' version

Dirbaio:
> not to mention the fact that the "IP versions" were integration tests?

ST version naming makes no sense whatsoever, don't try to read much into it :rofl: 

> Versions now match what is on spreadsheet

The spreadsheet is not necessarily correct either. The only reliable source is the ST reference manuals. Have you checked against them, for every family, to verify your changes are correct?

TaylorCarpenter:
> > Versions now match what is on spreadsheet
> 
> 
> 
> The spreadsheet is not necessarily correct either. The only reliable source is the ST reference manuals. Have you checked against them, for every family, to verify your changes are correct?

I spot checked 4 or 5. I'll check the rest of them now. The ST site was being particularly slow with downloading the reference manuals when I was working on this before.

Dirbaio:
just wanted to check you hadn't gone just by the spreadsheet (that's maintained manually and usually not updated after a mistake is fixed in this repo). 

BTW the SVDs are also a reliable source, if you do `./d extract-all CRC` and check the extracted yamls that's usually reliable too. (see readme for how to add a peripheral). It's faster than checking the RMs.

Dirbaio:
> There are some chips that weren't explicitly called out in the spreadsheet, so we can fall back on the original checks against 'integtest' version

Could you change everything to use per-family versions? Mixing both "family" and "cubedb" version is confusing, and not that useful now that the cubedb version matching has been shown to be wrong. 

Seems all the newer chips are CRC v3.

TaylorCarpenter:
> > There are some chips that weren't explicitly called out in the spreadsheet, so we can fall back on the original checks against 'integtest' version
> 
> Could you change everything to use per-family versions? Mixing both "family" and "cubedb" version is confusing, and not that useful now that the cubedb version matching has been shown to be wrong.
> 
> Seems all the newer chips are CRC v3.

Done.

Dirbaio:
thanks!

PR #375, opened 2024-01-30
Add HASH

caleb-garrett:
Adds support for HASH peripheral on all chips.


Dirbaio:
Can you remove fieldsets that are a single 32-bit wide field? such as HR, DIN, CSR... these don't give much information and make using the PAC a bit more annoying, it's easier if the entire register is a raw u32.

Dirbaio (approved):
thank you!

PR #377, opened 2024-01-31
[READY NOW!] Initial Comparator Support

AdinAck:
Add comparator support for G0 and G4.


stm32-data-gen/src/chips.rs
```diff
@@ -536,6 +536,8 @@ impl PeriMatcher {
             (".*:CRYP:cryp1_v1_0.*", ("cryp", "v1", "CRYP")),
             (".*:CRYP:cryp1_v2_0.*", ("cryp", "v2", "CRYP")),
             (".*:CRYP:cryp1_v2_2.*", ("cryp", "v2", "CRYP")),
+            ("STM32G0.*1.*:.*:COMP:.*", ("comp", "v1", "COMP")),
```
eZioPan, line 539: If you mean "filter G0x1", then it’s better to use "STM32G0?1.*"?

AdinAck: Haha yep I do mean that, I am *not* good at regex lol, just tried copying what was above.

AdinAck: Hmm, didn't seem to work when I pasted your suggestion in. I think i'll leave it alone for now.

AdinAck: Oh I think you meant `.` instead of `?`, will update, thanks.

eZioPan: Oh sorry I made a mistake, it’s "STM32G0.1"

lucasgranberg:
I am also looking in to implementing COMP but for WL series. Both WL and L4 series need a v3 so outside of the scope of this PR. (Difference in 20:18 BLANKING vs 21:19 BLANKSEL)

INMSEL and INPSEL could be done with enums but probably smart to keep it as u8 as every chip will probably have different versions of this.

 In my case COMP1 and COMP2 also differs as COMP2 has a WINMODE flag (windowing).

AdinAck:
Yeah at first I had enums for everything, and as such multiple COMP versions per chip. @Dirbaio told me why that was not a good idea, and to make those values integers so a HAL implementation would be much easier and macroless :)

lucasgranberg:
I have my stm32wl version working well enough. Needs a v3 but that is easy to add after this PR is done/merged. What needs to happen before this is ready?

lucasgranberg:
allright! maybe remove the [WIP] then :+1:

AdinAck:
Haha well I meant WIP as in there are like 20 *more* comparators that will need to be added later, but maybe you're right, I have changed the name :)

Dirbaio (approved):

PR #378, opened 2024-02-04
Corrected hash STR register to RW

caleb-garrett:
The hash STR register is typically write-only, but for context-restoring use it is necessary to be able to read this register. This PR marks the STR register in the v1 config as read/write, matching the other configs.


Dirbaio:
LGTM. please don't autoformat the YAML files though.

caleb-garrett:
Apologies that slipped past me. Auto-format undone.

Dirbaio (approved):
thanks!

PR #381, opened 2024-02-07
adds adc support for L0

shufps:
The L0 ADC is mostly compatible to `v1` but with additional registers and it also has some rather smaller differences (like `SAMPLE_TIME`)

This is part one of the PR, part two for `embassy-rs` that uses the `l0`-peripherals will come a little later (but I tested it successfully already).


Dirbaio (approved):
thanks!

PR #382, opened 2024-02-07
add comp_v3 and apply to stm32wle

lucasgranberg:
COMP1 and COMP2 on stm32wl differs only on the WINMODE bit for COMP2. I opted to skip that bit.


lucasgranberg:
L4 should have the same core but there are some differences in the register layouts (WL has INMESEL). guessing the wl is the odd one out as usual.

STM32WLE
image
STM32L4
image

Dirbaio:
thanks :crab: :rocket:

PR #387, opened 2024-02-07
Add OCTOSPI1 register bits

### LINKED ISSUE #384: STM32H723ZG is missing implementation of RccPeripheral for OCTOSPI1
The `OCTOSPI2` peripheral has an rcc config, but the `OCTOSPI1` is missing rcc information.

msrd0:
Those were previously only called QUADSPI, which is needed on some chips like the STM32H745, but those bits are used as OCTOSPI1 bits on other chips like the STM32H723.

Should fix #384


Dirbaio:
thanks!

PR #388, opened 2024-02-07
Fix register address for OCTOSPI2 peripheral

msrd0:
This fixes the address for the STM32H723 chips. It also updates a lot of other chips, hopefully those were incorrect before as well.


Dirbaio:
thanks!

PR #394, opened 2024-02-15
remove clippy warning and error from genrated files.

eZioPan:
combine with embassy-rs/chiptool#23, it removes *1k+* clippy warning and error from generated source code, with "stm32h573zi","pac","metadata" feature enabled.


Dirbaio:
Thanks!

PR #395, opened 2024-02-19
forbid enable multiple stm32 features ...

eZioPan:
... when use stm32-metapac directly.

the code is copied from `embassy-stm32/build.rs`


Dirbaio (approved):
nice!

PR #401, opened 2024-02-21
icache

eZioPan:


data/registers/icache_v1_4crr.yaml
```diff
@@ -0,0 +1,204 @@
...
+  - name: ERRF
+    description: an error occurred during the operation.
+    bit_offset: 2
+    bit_size: 1
+enum/CACHEINV:
+  bit_size: 1
+  variants:
+  - name: Invalidate
+    description: Invalidate entire cache
+    value: 1
```
Dirbaio, line 152: please remove these "write 1 to do thing" enums, they're not useful. Same as the 0=disabled, 1=enabled enums. it's better for the field to be a bool in the PAC.

Dirbaio (approved):

PR #407, opened 2024-02-22
ADC support for H5

tcbennun:


tcbennun:
Temperature reads tested. See https://github.com/embassy-rs/embassy/pull/2613

Dirbaio (approved):
thanks!

PR #409, opened 2024-02-23
adc h5 improving

eZioPan:
do a format, add a few enums and fields, recreate transform yaml files.

by the way, I use a feature from embassy-rs/chiptool#26 to offset registers' byte_offsets of `ADC_COMMON` block.
Please consider merge that first before merge this PR, thanks!


Dirbaio:
thanks!

PR #411, opened 2024-02-23
H7 RCC CRYP Register Naming

caleb-garrett:
The trait `rcc::RccPeripheral` is not implemented for CRYP on the H7 due to an apparent naming mismatch. This just changes the name for all registers `CRYPT` to `CRYP`.


Dirbaio (approved):
thanks!

PR #428, opened 2024-02-27
opamp for H7 and H5

eZioPan:
after this, all peripherals of all H5 chips get mapped.


Dirbaio:
> all peripherals of all H5 chips get mapped.

:partying_face: !!!!

thanks for your work, this is great!

PR #433, opened 2024-03-04
fix ADC, DAC clock muxes for H5, U5

tcbennun:
the clock selection bit is named ADCDACSEL, shared between all ADCs and DACs

### DIFF
--- a/data/registers/rcc_h5.yaml
+++ b/data/registers/rcc_h5.yaml
@@ -2114,7 +2114,7 @@ fieldset/SECCFGR:
 enum/ADCDACSEL:
   bit_size: 3
   variants:
-  - name: HCLK1
+  - name: HCLK2
     description: rcc_hclk selected as kernel clock (default after reset)
     value: 0
   - name: SYS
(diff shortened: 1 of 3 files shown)


tcbennun:
Tested with ADC1 on H503. The `ClockMux::adcdacsel` field plus init code is generated correctly

Dirbaio:
thanks!

(the u5 exception is a bit unfortunate, but perhaps it's our "fault" because all HCLKs are the same freq (almost always :smiling_face_with_tear:) , so perhaps we should be using HCLK instead of HCLK1/HCLK2/etc.)

tcbennun:
> thanks!
> 
> (the u5 exception is a bit unfortunate, but perhaps it's our "fault" because all HCLKs are the same freq (almost always 🥲) , so perhaps we should be using HCLK instead of HCLK1/HCLK2/etc.)

yeah; for APB periphs it makes sense as you have a prescaler per APB clock, but AFAIK all AHBs will share the same prescaler... do you know some exceptions to this?

Dirbaio:
stm32wl/stm32wb have separate prescalers for core2_ahb, shared_ahb.
stm32wba has a separate prescaler for ahb5.

Issue #444, opened 2024-04-03
HSEM peripheral missing in cube xml files for several chip families


taunusflieger:
I'm in the process to add support for HSEM. 

It looks like that the cube XML files are missing the HSEM entry except for the STM32WLE family. `parse_groups()` relays on the xml data to identify the IP available on the chip. If HSEM is not defined in the XML the further process does not process the HSEM data. Therefore the data generation process is only successful for the STM32WEL family. For all other chips which have a HSEM peripheral the data generation process is not successful. 

I've done some quick hack to prove that the missing HSEM entry in the cube XML is causing the issue:

```
fn process_group(
    mut group: ChipGroup,
    peri_matcher: &mut PeriMatcher,
    headers: &header::Headers,
    af: &gpio_af::Af,
    chip_interrupts: &interrupts::ChipInterrupts,
    peripheral_to_clock: &rcc::ParsedRccs,
    dma_channels: &dma::DmaChannels,
    chips: &HashMap<String, Chip>,
    memories: &memory::Memories,
    docs: &docs::Docs,
) -> Result<(), anyhow::Error> {
    let chip_name = group.chip_names[0].clone();
    group.family = Some(group.xml.family.clone());
    group.line = Some(group.xml.line.clone());
    group.die = Some(group.xml.die.clone());
    let rcc_kind = group.ips.values().find(|x| x.name == "RCC").unwrap().version.clone();
    let rcc_block = peri_matcher
        .match_peri(&format!("{chip_name}:RCC:{rcc_kind}"))
        .unwrap_or_else(|| panic!("could not get rcc for {}", &chip_name));
    let h = headers
        .get_for_chip(&chip_name)
        .unwrap_or_else(|| panic!("could not get header for {}", &chip_name));
    let chip_af = &group.ips.values().find(|x| x.name == "GPIO").unwrap().version;
    let chip_af = chip_af.strip_suffix("_gpio_v1_0").unwrap();
    let chip_af = af.0.get(chip_af);

    // Quick hack to add the missing HSEM IP
    if chip_name.starts_with("STM32H747") || chip_name.starts_with("STM32H757") {
        info!("Patching HSEM in IPS for {}", chip_name);
        group.ips.insert(
            "HSEM".to_string(),
            xml::Ip {
                name: "HSEM".to_string(),
                version: "hsem1_v1_0_Cube".to_string(),
                instance_name: "HSEM".to_string(),
            },
        );
    }
```

What is the right approach to get this fixed?

PR #450, opened 2024-04-04
Add support for HSEM

taunusflieger:
Add support for the hardware semaphore 

The main issue with implementing HSEM is that the Cube XML is missing HSEM information except for the wle5. Without the Cube XML information the data generation process will not generate the data for HSEM. Therefore a preliminary patch has been implemented in chips.rs (see also https://github.com/embassy-rs/stm32-data/issues/444). I'm unsure about the implemented patch - would appriciate feedback.

chips with HSEM 
h745
h747
h755
h757
h735
h7b3
h753v
h753
h743
h743v
mp157
mp153
wb55
wba52
wl5x_cm0p
wl5x_cm4
wle5 

The current draft PR contains the HSEM implementation for: 

h745
h747
h755
h757
wle5


taunusflieger:
Hey @Dirbaio would appreciate your input here

Dirbaio:
We already have [GHOST_PERIS](https://github.com/embassy-rs/stm32-data/blob/2b7ec569a5510c324693f0515ac8ea20b12917a9/stm32-data-gen/src/chips.rs#L1006) for handling peripherals that are missing in cubedb, can you add HSEM there instead of adding custom code?

Also, there's no way there's 8 versions. from looking at the RMs, there's:
  - h7 singlecore: 32 semaphores, 1 irq line.
  - h7 multicore: 32 semaphores, 2 irq lines (one per core)
  - wb: 32 semaphores, 2 irq lines (one per core)
  - wba: 16 semaphores, trustzone shit (LOCK, PRIV bits, secure and nonsecure irq lines)
  - wl5: 16 semaphores, 2 irq lines (one per core)
  - wle: 16 semaphores, 1 irq line

H7 multicore and WB are the same, so this is 5 versions in total.

Can you rearrange versions to match that? also, please make sure to cleanup the yamls (convert stuff to arrays, remove dupe enums, remove useless single-bot enums...), and minimize the diff between versions for easy review.

taunusflieger:
The following list shows the different hsem yaml versions and the corresponding chips.
wba is on purpose not included at is complex and very different from the others which will also make the HSEM implementation in the HAL more complex. I leave this out for another PR.

**v1:**
- h747
- wb55

**v2:**
- h735
- h7b3
- h753v
- h753
- h743
- h743v

**v3:**
- wl5x_cm0p
- wl5x_cm4

**v4:**
- wle5

Dirbaio:
- in v1,v2 coreid is 4 bits, not 8bits. in CR, R, RLR.
- To make it easier to make the HAL to work on all chips, you might want to unify the registers more. Up to you, though. Here are some ideas:
  - rename "MASTERID" to "COREID"
  - Always make the interrupt registers an array, just make it length 1 or 2 depending on the chip.

Dirbaio (approved):
thanks!

PR #451, opened 2024-04-04
lptim v2

eZioPan:
this PR cover

l5 wl as lptim_v2a
h5 u5 wba as lptim_v2b


data/registers/lptim_v2b.yaml
```diff
@@ -0,0 +1,728 @@
...
+    fieldset: ICR_IC
+  - name: DIER
+    description: LPTIM interrupt enable register.
+    byte_offset: 8
+    fieldset: DIER_IC
+block/LPTIM_ADV:
+  extends: LPTIM_BASIC
+  description: Low power timer with Output Compare
+  items:
+  - name: InputCapture
```
Dirbaio, line 19: Instead of doing two overlapping blocks for the two modes like the RM, could you merge the bits in the two modes of the registers? it makes it simpler. 

I don't know why the RM splits it like this, all the bits that are common between the two modes match exactly, there's no "this bit does X in capture mode and Y in compare mode"....

PR #455, opened 2024-04-05
Add 8 and 16 bit wide SPI data registers

dlaw:
Most SPI peripherals are sensitive to the bit width of reads/writes to DR/TXDR/RXDR. This PR adds 8 and 16 bit wide versions of those registers to permit such accesses to be generated. This is the same approach already used for the [CRC DR](https://github.com/embassy-rs/stm32-data/blob/main/data/registers/crc_v2.yaml#L7) in crc_v2.yaml and crc_v3.yaml.

* spi_v1: This version is not sensitive to the bit width of the read or write. Only the low 8 bits are used.

* spi_v2: This version is sensitive to the bit width of the read or write. Data packing is automatically performed based on the width of the access. From the STM32G4 reference manual:
2024-04-05-133820_863x359_scrot

* spi_v3, spi_v4, spi_v5: These versions are sensitive to the bit width of the read or write. From the STM32U5 reference manual:
2024-04-05-132902_633x407_scrot

### DIFF
--- a/data/registers/crc_v2.yaml
+++ b/data/registers/crc_v2.yaml
@@ -1,7 +1,7 @@
 block/CRC:
   description: Cyclic Redundancy Check calculation unit
   items:
-  - name: DR
+  - name: DR32
     description: Data register
     byte_offset: 0
   - name: DR16
--- a/data/registers/crc_v3.yaml
+++ b/data/registers/crc_v3.yaml
@@ -1,7 +1,7 @@
 block/CRC:
   description: Cyclic Redundancy Check calculation unit
   items:
-  - name: DR
+  - name: DR32
     description: Data register
     byte_offset: 0
   - name: DR16
--- a/data/registers/spi_v2.yaml
+++ b/data/registers/spi_v2.yaml
@@ -13,10 +13,14 @@ block/SPI:
     description: status register
     byte_offset: 8
     fieldset: SR
-  - name: DR
-    description: data register
+  - name: DR16
+    description: data register - half-word sized
     byte_offset: 12
-    fieldset: DR
+    bit_size: 16
+  - name: DR8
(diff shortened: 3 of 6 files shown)


Dirbaio:
oh yes this is a *very* good idea. Can you rename `DR` to `DR32` for consistency?

dlaw:
Sure, a few questions:
1. Do you want DR renamed to DR32 for the CRC also? Currently CRC has DR (32 bit), DR16 (16 bit), DR8 (8 bit).
2. spi_v2 doesn't support a 32 bit mode. Its 32-bit DR is currently defined to have a single 16 bit field. The reference manual actually lists DR as a 16 bit wide register. For this case should I delete DR32 from the peripheral?
2024-04-05-135148_988x471_scrot

Dirbaio:
> Do you want DR renamed to DR32 for the CRC also? Currently CRC has DR (32 bit), DR16 (16 bit), DR8 (8 bit).

hmm that's true, we should rename that for consistency too. so why not

> spi_v2 doesn't support a 32 bit mode. Its 32-bit DR is currently defined to have a single 16 bit field. The reference manual actually lists DR as a 16 bit wide register. Should I delete DR32 from spi_v2?

I guess yes? if writing DR32 pushes just 16bits to the fifo, same as writing DR16, there's no reason for DR32 to exist I think?

> I kept the 32-bit DR for backwards compatibility, but if we're breaking it with a name change anyway...

in general we don't worry about compat, every stm32-metapac release is breaking anyway because we fix some random mistake or rearrange registers somewhere.

dlaw:
Changes pushed. I got rid of the trivial fieldsets, so all the widths of CRC DRs and SPI DRs are now consistent in that regard.

Dirbaio (approved):
thanks!

Issue #301, opened 2023-11-01
Incorrect memory sizes tracking issue


Dirbaio:
If you've got a chip where the `stm32-metapac/memory.x` feature generates a wrong memory.x please post here.

Identified cases:

- STM32G030K8 flash generates 32kb but has 64kb
- STM32WL55JC ram generates 12kb but has 64kb

TaylorCarpenter:
STM32WB55VG ram generates 128kb but has 192kb.

phycrax:
STM32G030C8 flash generates 32kb but has 64kb

trepidacious:
stm32g491re memory.x shows 32kb RAM, chip should have 112kb (3 blocks of 80, 16, 16)

Dirbaio:
STM32F303ZE says it has 64k but only has 48kb https://github.com/embassy-rs/stm32-data-generated/issues/1
stm32u5a5zj says it has 32k but it has 2.5mb or something

eZioPan:
STM32F411RE:

| Type | memory.x | DS Tab.2  | Note |
|------| -------- | --------- | ---- |
| RAM  |   64K    |  128 KB   |      |

eZioPan:
|     Chip         |  Type  | memory.x | DS Tab.2               |        Note        |
| :--------------: | :----: | :------: | :--------------------: | :----------------: |
|  STM32H5[67]xxG  |  RAM   |  256 K   |  640 (256+64+320) KB   |                    |
|  STM32H5[67]xxI  |  RAM   |  256 K   |  640 (256+64+320) KB   |                    |

Dirbaio:
|     Chip    |  Type  | memory.x | actual  |        Note        |
| :---------: | :----: | :------: | :-------: | :----------------: |
| STM32F051R8T6 |  RAM   |  7 K   |  8 KB   |   #437         |

LorenzoCucchi:
|     Chip    | Type | memory.x | actual | Note |
|:-----------:|:----:|:--------:|:------:|:----:|
| STM32F756ZG |  RAM |   320K   |  256K  |      |

64 K less since DTCM starting at 0x20000000 is not used, and RAM starts at 0x20010000.

adri326:
|     Chip    |  Type  | memory.x | DS Tab.2  |        Note        |
| :---------: | :----: | :------: | :-------: | :----------------: |
| STM32L4A6ZG |  RAM   |  320K   |  256K   |                    |

[...]

PR #460, opened 2024-04-09
Manually maintain memory maps instead of parsing them from cubeprogdb.

Dirbaio:
- **Remove OTP from memory map.**
- **Sort memory regions  by addr.**
- **Manually maintain memory maps instead of parsing them from cubeprogdb.**

First step towards fixing #301

The cubeprogdb has turned out to be a quite bad data source. It's not granular
enough (it has one entry per chip die, not per chip) so the previous code joined
the data with the C headers and cubedb to fill in the gaps, essentialy "guessing"
stuff. This has been quite error prone (see #301) and hard to make fixes to.

Instead, we're going to manually maintain memory maps in a .rs file. This way, if
something is wrong we can simply go and fix it.

This commit just migrates the existing data, even if it's wrong. (it does fix
a few very minor mistakes). Next steps is actually fixing the memory maps.

### DIFF
--- a/stm32-data-gen/src/chips.rs
+++ b/stm32-data-gen/src/chips.rs
@@ -1356,7 +1356,6 @@ fn process_chip(
         "FLASH_BANK1",
         "FLASH_BANK2",
         "FLASH",
-        "FLASH_OTP",
         "D1_AXIFLASH",
         "D1_AXIICP",
     ] {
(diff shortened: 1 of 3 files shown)


rmja:
It is a shame that we lost the otp information and that https://github.com/embassy-rs/embassy/commit/1f3b690f76b6732e1ff152091040cec833da25cf was merged. What is the proper way to restore this functionality?

Issue #438, opened 2024-03-15
STM32G4 should have two copies of ADC_COMMON


dlaw:
All instances of STM32G4 should have `ADC12_COMMON` at address `0x50000300` in addition to `ADC345_COMMON` at address `0x50000700`. Currently stm32-metapac provides `ADC_COMMON` for ADC1/2 and does not map any of the common registers for ADC3/4/5.

![2024-03-15-170218_632x579_scrot](https://github.com/embassy-rs/stm32-data/assets/606010/e98ec8e9-c167-4530-a818-ea6fb2ca11f7)

![2024-03-15-170231_644x218_scrot](https://github.com/embassy-rs/stm32-data/assets/606010/6ba4bd0c-fc5d-46b6-a7db-ae40b5e5d84a)

dlaw:
It's not clear to me how to fix this the right way. I can add ADC345_COMMON to data/extra/family/STM32G4.yaml, no difficulty there, but now I can't figure out how to rename the existing ADC_COMMON to ADC12_COMMON.

PR #461, opened 2024-04-09
Add the missing ADC3_COMMON peripheral for STM32G4

dlaw:
This PR addresses https://github.com/embassy-rs/stm32-data/issues/438 by adding a new ADC345_COMMON peripheral for the STM32G4.

*Question for maintainers:* Should the existing ADC_COMMON peripheral on the G4 be renamed to ADC12_COMMON for consistency? If so, where is the right place to make this change?

### DIFF
--- a/stm32-data-gen/src/chips.rs
+++ b/stm32-data-gen/src/chips.rs
@@ -213,24 +213,21 @@ impl PeriMatcher {
             ("STM32WL5.*:ADC:.*", ("adc", "g0", "ADC")),
             ("STM32WLE.*:ADC:.*", ("adc", "g0", "ADC")),
             ("STM32G0.*:ADC:.*", ("adc", "g0", "ADC")),
-            ("STM32G0.*:ADC_COMMON:.*", ("adccommon", "v3", "ADC_COMMON")),
             ("STM32U0.*:ADC:.*", ("adc", "u0", "ADC")),
-            ("STM32U0.*:ADC_COMMON:.*", ("adccommon", "v3", "ADC_COMMON")),
             ("STM32G4.*:ADC:.*", ("adc", "g4", "ADC")),
+            ("STM32G0.*:ADC_COMMON:.*", ("adccommon", "v3", "ADC_COMMON")),
+            ("STM32U0.*:ADC_COMMON:.*", ("adccommon", "v3", "ADC_COMMON")),
             ("STM32G4.*:ADC_COMMON:.*", ("adccommon", "v4", "ADC_COMMON")),
             (".*:ADC_COMMON:aditf2_v1_1", ("adccommon", "v2", "ADC_COMMON")),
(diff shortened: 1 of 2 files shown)


dlaw:
This change has the desired effect for STM32G4, but it looks like it is also adding an empty (no registers) ADC12_COMMON peripheral for a bunch of other chips. Is this a problem? I checked a few those chips, and it looks like (1) they don't have any ADC_COMMON hardware, (2) each one already had an empty (no registers) peripheral called ADC_COMMON.

Dirbaio:
that might be because there's some code that ignores numbers when matching peripherals to C headers... you'd have to hunt that down and add exceptions for this or improve it.

dlaw:
Okay, I'll try to untangle what's happening here.

Can you confirm whether the pre-existing ADC_COMMON peripherals lacking register maps (e.g. for STM32F103) are erroneous? I'll look at removing those also.

dlaw:
Okay, I took quite a journey through F3 and H7 datasheets, and I think I understand the situation now.

For cross-series consistency between F3, H7, and G4, the missing ADC_COMMON peripheral for G4 should be called ADC3_COMMON.

The pre-existing issue with an empty ADC_COMMON being generated for e.g. STM32F1 is unchanged by this PR.

dlaw:
@Dirbaio Thanks for the guidance. Here is a revised commit which introduces the correct names for G4.

This commit _should_ make it easy to introduce the correct naming for F3 and H7 also. However, I found that when I made that seemingly straightforward change, the ADC_COMMON peripheral disappeared entirely from many of the STM32F301 and STM32F302 parts. See https://github.com/dlaw/stm32-data/commit/broken which results in a diff like the following:
```
diff -r chips_orig/STM32F301C6.json build/data/chips/STM32F301C6.json
364,372d363
<                     "name": "ADC_COMMON",
<                     "address": 1342178048,
<                     "registers": {
<                         "kind": "adccommon",
<                         "version": "f3",
<                         "block": "ADC_COMMON"
<                     }
<                 },
<                 {
diff -r chips_orig/STM32F301C8.json build/data/chips/STM32F301C8.json
374,382d373
<                     "name": "ADC_COMMON",
<                     "address": 1342178048,
<                     "registers": {
<                         "kind": "adccommon",
<                         "version": "f3",
<                         "block": "ADC_COMMON"
<                     }
<                 },
<                 {
diff -r chips_orig/STM32F301K6.json build/data/chips/STM32F301K6.json
360,368d359
<                     "name": "ADC_COMMON",
<                     "address": 1342178048,
<                     "registers": {
<                         "kind": "adccommon",
<                         "version": "f3",
<                         "block": "ADC_COMMON"
<                     }
<                 },
<                 {
...
```

If you have some idea of the issue with (some of) the F3 parts when renaming ADC_COMMON to ADC12_COMMON, please let me know. I am completely mystified.

Otherwise, it would be good to merge this PR now to get G4 correct, and figure out the other issue later.

Dirbaio:
fixed in #475

PR #468, opened 2024-04-15
Add DSIHOST support

JuliDi:
This PR adds support for DSIHOST to stm32-data.

The yaml files were generated by using `./d extract-all DSI` and `./d extract-all DSIHOST`, which seems to have caught everything.
This resulted in YAML files for:
- DSI:
  - f469
  - f765 (Actually wrong, f765 does not have DSI)
  - f7x7 (Actually wrong, f7x7 does not have DSI)
  - f7x9
  - l4r9
  - mp157
  - u599
  - u5a9
and
- DSIHOST:
  - h747cm4
  - h747cm7

Comparing the YAML files showed:
- h747cm4 and cm7 are 100% identical
- f7x7 and f7x9 are 100% identical (whatever that means)
- u599 and  u5a9 are 100% identical

There seem to be 3 different IP versions (Cube and SHARK_Cube appear to have no differences):
- `dsihost1_v1_0_Cube` is used by
    - STM32F469
    - STM32F479
    - STM32F768AIYx
    - STM32F769
    - STM32F778AIYx
    - STM32F779AIYx
    - STM32L4R9
    - STM32L4S9
- `dsihost1_v1_0_SHARK_Cube` is used by
    - STM32H747
    - STM32H757 
- `dsihost1_v2_0` is used by
    - STM32MP157
- `dsihost_U5` is used by
    - STM32U599
    - STM32U5A9
    - STM32U5F9
    - STM32U5G9

Accordingly, we have added `dsihost_v1.yaml`, `dsihost_v2.yaml` and `dsihost_u5.yaml`. However, so far only `dsihost_v1.yaml` is somewhat verified (having an STM32F469 DK at hand). The V2 is only used in the STM32MP157, for which embassy does not yet have support. But it probably does not hurt to have it in there already.

In perimap, the YAML files are now mapped like this, which seems to work fine:
```rust
(".*:DSIHOST:dsihost1_v1_0", ("dsihost", "v1", "DSIHOST")),
(".*:DSIHOST:dsihost1_v1_0_SHARK", ("dsihost", "v1", "DSIHOST")),
(".*:DSIHOST:dsihost1_v2_0", ("dsihost", "v2", "DSIHOST")),
(".*:DSIHOST:dsihost_U5", ("dsihost", "u5", "DSIHOST")),
```

Some minor changes to `header.rs` and `rcc.rs` to fix mixed use of `DSI` and `DSIHOST` were necessary.

We have decided to name the peripheral `DSIHOST` because this appears more often in the XML files and the "official" name in the Reference Manuals is "DSI Host" as well.

A PR to embassy that fixes issues with the RCC clock setup introduced with this PR (fix has been discussed in the chat) will follow shortly. If possible, this PR could be merged first such that the embassy PR can reference this metapac build already.


Dirbaio (approved):
thank you!

PR #476, opened 2024-04-30
render multi-line document properly

eZioPan:
and some clippy fix.

---

will need a 

- [ ] Accept embassy-rs/chiptool#31
- [ ] Rebase to remove my fork from codebase.

---

This PR doesn't change function of stm32-data-gen, so there should be no change in ci diff.

---

BELOW IS OUT OF DATE, SEE embassy-rs/chiptool#31 INSTEAD

---

Render metapac document from one-line to multi-line.

eg.

from

```rust stm32-metapac/src/peripherals/cordic_v1.rs
#[doc = "Scaling factor. Input value has been multiplied by 2^(-n) before for argument. Output value will need to be multiplied by 2^n later for results."]
```

to

```rust
#[doc = "Scaling factor.  "]
#[doc = "Input value has been multiplied by 2^(-n) before for argument.  "]
#[doc = "Output value will need to be multiplied by 2^n later for results."]
```

after

---

### DIFF
--- a/stm32-data-gen/src/chips.rs
+++ b/stm32-data-gen/src/chips.rs
@@ -114,6 +114,7 @@ struct PeriMatcher {
 
 impl PeriMatcher {
     fn new() -> Self {
+        #[rustfmt::skip]
         const PERIMAP: &[(&str, (&str, &str, &str))] = &[
             (".*:USART:sci2_v1_1", ("usart", "v1", "USART")),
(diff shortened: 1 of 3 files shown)


leftger:
This PR is more than 2 years old. As far as I can tell there's very little interest in getting multiline rendering properly for generated PACs. The metapac has changed substantially since then and I think the changeset would be much larger. If there is a streamlined proposal to accomplish the same result efficiently I'm sure a PR can be submitted for it.

PR #478, opened 2024-05-13
Fix Memory Size for STM32G4...6/8 and STM32G4[34]..B

AdinAck:
The memory size for these chips was set to `20` KiB instead of `32`. Perhaps a hex to dec conversion error.

Evidence that the correct size is `32` KiB:
[STM32G4xxxx](https://www.st.com/resource/en/datasheet/stm32g431c6.pdf)
[STM32G441RB](https://www.st.com/resource/en/datasheet/stm32g441rb.pdf)

Thanks!


AdinAck:
After looking closer at the datasheet I actually don't know how to modify the invocation of `mem!` to accurately represent the SRAM banks in these chips, but `20` KiB is still certainly wrong so I will change this to an issue.

AdinAck:
woops: #301

Dirbaio:
thanks!

PR #481, opened 2024-05-15
Added SPI45SEL to RCC for STM32H7 devices.

liarokapisv:
RCC mux was missing the SPI45SEL selector. I had to add an extra rule to the rcc.rs and fix the wrong SPI45SEL entries in rcc_h7ab.yaml and rcc_h7rs.yaml. I confirmed that both default entries point to the PCLK2 instead of HCLK2.


Dirbaio (approved):
thank you!

PR #482, opened 2024-05-20
added I2SCFGR register to spi_v3.yaml

liarokapisv:
The `I2SCFGR` register was missing from the `spi_v3.yaml` description. I double-checked that the register is the same between the `spi2s2_v1_0` and `spi2s2_v1_1` variants by diffing the svd files. Also fixed some `svd` typos in the process


liarokapisv:
I will try to fix some naming incompatibilities with the other spi versions so that we have fewer conflicts in the common drivers.

liarokapisv:
I think this is mostly it, I have a few reservations about changing the `I2SCFGR` enum itself, any feedback is welcome.

Dirbaio:
Have you checked that all SPIv3 chips have I2SCFGR? iirc newer chips with SPIv3 that have SAI have dropped the I2S stuff.

liarokapisv:
Two H7s (H743 & H745) that I have used, definitely have SAI, but also have I2S on their SPI because it is simpler (and can be used with SPI pins). In `stm32-data-gen/src/chips`, spi_v3 matches the `spi2s2_v1_0` and `spi2s2_v1_1` spi versions. I checked a subset of SVD files with those SPI versions, and the I2S stuff is definitely present in those. I also double-checked with reference manuals but again only for a subset with those versions.

Dirbaio (approved):
awesome! thanks for checking.

PR #483, opened 2024-05-29
Generate ALL_CHIPS and ALL_PERIPHERAL_VERSIONS metadata

honzasp:
As discussed in https://github.com/embassy-rs/embassy/pull/3005, let's export the list of all chip names and the list of all versions of all peripherals in the metadata, to be used in build.rs in embassy-stm32.


Dirbaio (approved):
thanks!

PR #484, opened 2024-06-05
Small timer fixes and rearrangements

honzasp:
- **Fix**: timer_v1 does not support field DBSS in register DCR
- Nitpick: in timer_v1, `block/TIM_2CH_CMP` does not need to override `DCR` and `DMAR`, because they are inherited from `block/TIM_1CH_CMP`
- Nitpick: in timer_v1, `fieldset/SMCR_ADV` was the same as `fieldset/SMCR_GP16`
- Nitpick: in timer_v1, `fieldset/DIER_ADV` does not need to override `BIE`, because it is inherited from `fieldset/DIER_1CH_CMP`
- **Fix**: in timer_v2, `block/TIM_2CH_CMP` should define `CCR_DITHER` as an array of len 2 (same as `CCR`)
- **Fix**: in timer_v2, `block/TIM_ADV` should define `ECR`
- Nitpick: in timer_v2, I put some registers in the same order as in timer_v1, to remove noise from the textual diff between timer_v1.yaml and timer_v2.yaml

I'm learning about the different types of STM32 timers and I created a [Google Sheet](https://docs.google.com/spreadsheets/d/1eeiW_iYWhJE245WIJOh_zEd9_FOLs4A49rqMteGU7Eo/edit?usp=sharing) with an overview of the various registers and their fields for different timer types and versions (only `v1` and `v2` at the moment, but I also want to have a look at `l0`). I found these fixes as irregularities in the table.

### DIFF
--- a/data/registers/timer_v1.yaml
+++ b/data/registers/timer_v1.yaml
@@ -46,15 +46,15 @@ block/TIM_1CH:
       stride: 4
     byte_offset: 52
     fieldset: CCR_1CH
+  - name: TISEL
+    description: input selection register
+    byte_offset: 104
+    fieldset: TISEL_1CH
   - name: OR
     description: |-
       Option register 1
       Note: Check Reference Manual to parse this register content
     byte_offset: 80
-  - name: TISEL
-    description: input selection register
-    byte_offset: 104
-    fieldset: TISEL_1CH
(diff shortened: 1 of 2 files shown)


Dirbaio (approved):
thanks!

PR #485, opened 2024-06-07
Further timer fixes

honzasp:
- **Fix**: in timer_v1, I removed `DCR` from `block/TIM_GP16` by mistake in #484 :see_no_evil:
- **Fix**: timer_l0 does not support field `DBSS` in register `DCR`
- **Fix**: in timer_v1 and timer_v2, register `DIER` of `TIM_2CH_CMP` has field `TIE`
- Nitpick: in timer_l0, `fieldset/SMCR_2CH` is the same as `fieldset/GP_16`


Dirbaio (approved):
thanks!

PR #486, opened 2024-06-08
Allow using stm32-data-gen as a library

bugadani:


Dirbaio:
why do you need this? the code is quite unstable, I wouldn't want to rely on it as a lib, it'll get broken quite often. Also it's not self-contained, it needs the `sources` repo to be cloned in the right (hardcoded) path, which is done from the `d` script, not from Rust.

If you need the output data for something I'd suggest consuming it from https://github.com/embassy-rs/stm32-data-generated/ . The format is much more stable.

bugadani:
Well this all happened while offline for 12 hours, so... didn't have metapac, went my own way. Does the metapac have an alternative to `memory::get`? I need to feed in a string and receive a memory layout, which was really convenient to do here.

Dirbaio:
i'd recommend reading the `memory` field from the json files in `stm32-data-generated` perhaps: https://github.com/embassy-rs/stm32-data-generated/blob/main/data/chips/STM32C011D6.json#L97

the `stm32-data-serde` crate in this repo has ready-to-use struct definitions if you want.

the info is also available on stm32-metapac but only for the currently enabled chip via Cargo features, if you want the info for all chips you'll have to build it 1400 times which is not what you want.

I'm reluctant to expose the bits in `stm32-data-gen` like in this PR because then that'll become public API, there'll be some expectation that we don't break it, which is a burden i'd prefer to not take. The "public API" of stm32-data is the generated json files and stm32-metapac, both available in the [stm32-data-generated](https://github.com/embassy-rs/stm32-data-generated) repo.

bugadani:
Well I'm happy with my "hack this PR into a local copy and run it once in a while". I don't need a minute-long script to parse a few hundred JSONs when I can call a function, even if I have to expose that function first.

PR #487, opened 2024-06-10
Add registers for SDADC

Stupremee:
This pull requests adds the register block for the Sigma-Delta ADC on the STM32F373 and STM32F301.
The documentation for the registers can be found in RM0313
I've never done a pull request to this repository, so I hope I've done everything correctly.

Before merging, there are a few open questions I have:

-  I want to remove the `{R,J}DATA{12,13}R` registers, because they only exist on the SDADC1 instance, and are not really necessary because they are just mirrors for the registers from other instances. Is there a way to do that via the transformations or can I do this manually?
    - Unfortunaly, if I extract the YAML files using the `./d extract-all SDADC{1,2,3}` script, all three instances contain these registers. That's why I kept these registers in the first place
    - Or maybe we should keep these registers, but only make them accessible in SDADC1?
- In the reference manual, there is a "Register Write protection" table for some registers, how is stuff like this handled inside this crate?
image

### DIFF
--- a/data/registers/sdadc_v1.yaml
+++ b/data/registers/sdadc_v1.yaml
@@ -1,0 +1,297 @@
+block/SDADC:
+  description: Sigma-delta analog-to-digital converter.
+  items:
+  - name: CR1
+    description: control register 1.
+    byte_offset: 0
+    fieldset: CR1
+  - name: CR2
+    description: control register 2.
+    byte_offset: 4
+    fieldset: CR2
+  - name: ISR
+    description: interrupt and status register.
+    byte_offset: 8
+    access: Read
+    fieldset: ISR
+  - name: CLRISR
+    description: interrupt and status clear register.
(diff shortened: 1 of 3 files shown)


Dirbaio:
> I want to remove the {R,J}DATA{12,13}R registers, because they only exist on the SDADC1 instance, and are not really necessary because they are just mirrors for the registers from other instances. Is there a way to do that via the transformations or can I do this manually?

what we usually do is just keep the registers. it's up to higher layers (the HAL) to know they can't access some.

> In the reference manual, there is a "Register Write protection" table for some registers, how is stuff like this handled inside this crate?

there's the `access` field on registers, but it can't dynamically protect based on the value of other regs. so let's just leave them as readwrite and leave higher layers to handle it too.

Dirbaio (approved):
thanks!

PR #494, opened 2024-06-18
Kkoppul2/u5 comp support

kkoppul2:
Add U5 comparator support
Clean up U0 comparator yaml


Dirbaio:
why remove the winout, winmode, polarity enums? they're 1-bit, but they're not "useless" because they convery meaning. For example "polarity: NotInverted, Inverted" makes sense, "polarity: true/false" doesn't. it'd make sense if the field was named "invert: true/false" but it isn't.

kkoppul2:
That's fair, I'll update that.

kkoppul2:
@Dirbaio Fixed it

Dirbaio (approved):
thanks!

PR #495, opened 2024-06-26
Refactor timer taxonomy

honzasp:
This PR accompanies https://github.com/embassy-rs/embassy/pull/3123. It simplifies the register definitions for timers, unifies the naming of timers and supports the type-safe timer taxonomy from that PR.

However, as a necessary trade-off, it exposes registers and register fields for all channels even for timers that don't have the full number of channels: for example, all four CCR registers are exposed for `Tim1ch`. The idea here is that `Tim1ch` represents a timer that is _at least as capable_ as a 1-channel timer, but might have up to 4 channels. For example, this allows us to implement a PWM driver that works for all timers with channels (it requires only the functionality of `Tim1ch`), but still supports up to 4 channels for timers that have them (so it can be used with `Tim4ch` or `TimAdv2ch`).

The idea is that the higher-level code ensures that only the available channels are used. For example, the `SimplePwm` driver ensures this by accessing only those channels that correspond to pins that belong to the timer peripheral: if there is a pin for channel 3, then the timer surely supports channel 3!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 /sources
 /tmp
 .idea/
+Session.vim
 .DS_Store
 # transform*.yaml
 __pycache__
--- a/data/registers/timer_l0.yaml
+++ b/data/registers/timer_l0.yaml
@@ -1,125 +1,5 @@
-block/TIM_1CH:
-  extends: TIM_CORE
-  description: Virtual 1-channel timers
-  items:
-  - name: CR1
-    description: control register 1
-    byte_offset: 0
-    bit_size: 16
-    fieldset: CR1_1CH
-  - name: DIER
-    description: DMA/Interrupt enable register
-    byte_offset: 12
-    fieldset: DIER_1CH
-  - name: SR
-    description: status register
-    byte_offset: 16
-    fieldset: SR_1CH
-  - name: EGR
-    description: event generation register
-    byte_offset: 20
-    access: Write
-    bit_size: 16
-    fieldset: EGR_1CH
-  - name: CCMR_Input
-    description: capture/compare mode register 1 (input mode)
-    array:
-      len: 1
-      stride: 4
-    byte_offset: 24
-    fieldset: CCMR_Input_1CH
-  - name: CCMR_Output
-    description: capture/compare mode register 1 (output mode)
-    array:
-      len: 1
-      stride: 4
-    byte_offset: 24
-    fieldset: CCMR_Output_1CH
-  - name: CCER
-    description: capture/compare enable register
-    byte_offset: 32
-    fieldset: CCER_1CH
-  - name: CCR
-    description: capture/compare register x (x=1)
-    array:
-      len: 1
-      stride: 4
-    byte_offset: 52
-    fieldset: CCR_1CH
-  - name: OR
-    description: |-
-      Option register 1
-      Note: Check Reference Manual to parse this register content
-    byte_offset: 80
-block/TIM_2CH:
-  extends: TIM_1CH
-  description: 2-channel timers
-  items:
-  - name: CR2
-    description: control register 2
-    byte_offset: 4
-    fieldset: CR2_2CH
-  - name: SMCR
-    description: slave mode control register
-    byte_offset: 8
-    fieldset: SMCR_GP16
-  - name: DIER
-    description: DMA/Interrupt enable register
-    byte_offset: 12
-    fieldset: DIER_2CH
-  - name: SR
-    description: status register
-    byte_offset: 16
-    fieldset: SR_2CH
-  - name: EGR
-    description: event generation register
-    byte_offset: 20
-    access: Write
-    bit_size: 16
-    fieldset: EGR_2CH
-  - name: CCMR_Input
-    description: capture/compare mode register 1 (input mode)
-    array:
-      len: 1
-      stride: 4
-    byte_offset: 24
-    fieldset: CCMR_Input_2CH
-  - name: CCMR_Output
-    description: capture/compare mode register 1 (output mode)
-    array:
-      len: 1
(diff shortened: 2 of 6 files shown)


honzasp:
Marked as draft because I want to double-check that I didn't introduce any unintentional breakage (in particular in the v2 and l0 versions).

honzasp:
This PR should now be ready for review :heavy_check_mark:

Dirbaio:
> However, as a necessary trade-off, it exposes registers and register fields for all channels even for timers that don't have the full number of channels: for example, all four CCR registers are exposed for Tim1ch. The idea here is that Tim1ch represents a timer that is at least as capable as a 1-channel timer, but might have up to 4 channels. For example, this allows us to implement a PWM driver that works for all timers with channels (it requires only the functionality of Tim1ch), but still supports up to 4 channels for timers that have them (so it can be used with Tim4ch or TimAdv2ch).

can't you already do this by unsafely casting a Tim1ch to a Tim4ch? For example, when you construct a SimplePwm, store it in `self` as a `Tim4ch` but record either in a field or in the type that it actually had only 1 ch, then avoid using the higher channels.

IMO this is better than changing the registers to allow 4ch in the 1ch timers, because:

There's 2 kinds of users of the PAC:
- HAL authors, writing generic code over all timers
- end users, using raw register access with one particular timer instance.

increasing the number of channels impacts end users (they can accidentally try using channels that don't exist), while requiring unsafe casting impacts only HAL authors, which are likely more "sophisticated" users.

honzasp:
> can't you already do this by unsafely casting a Tim1ch to a Tim4ch? For example, when you construct a SimplePwm, store it in `self` as a `Tim4ch` but record either in a field or in the type that it actually had only 1 ch, then avoid using the higher channels.

Sure, it's possible to treat any general-purpose timer as a `Tim4ch`, but the disadvantage is that `Tim4ch` gives you access to many registers and fields that are not available in `Tim1ch` and `Tim2ch`.

> increasing the number of channels impacts end users (they can accidentally try using channels that don't exist), while requiring unsafe casting impacts only HAL authors, which are likely more "sophisticated" users.

I agree that this change makes life slightly harder for people who use this crate directly, but it should make life significantly easier for people who use Embassy.

In the current situation, we protect end users of the PAC crate from using non-existent channels in a timer (i.e., if they try to use channel 3 in a 2-channel timer, they get a compile-time or a runtime error), but we don't protect end users of Embassy from using the wrong timer type (e.g., if they try to use `CountingMode::EdgeAlignedDown` with a 2-channel timer, they get no error).

With the proposed changes, we won't be able to protect the users of the PAC from using more channels, but we will be able to ensure that users of Embassy use timer peripherals that actually support all operations that they need.

In my opinion, it's a good trade-off: it's quite simple to understand that a 2-channel timer has only two channels, but understanding all types of STM32 timers and their features takes a few days of detailed studying of reference manuals.

honzasp:
Hi @Dirbaio, I got back to programming STM32 recently, and I again hit the issue that I tried to solve with this PR, this time as a direct user of the PAC: writing generic drivers for STM32 timers is challenging, because the hierarchy of timer capabilities is complex and not explicitly documented anywhere, and the only way how to write a generic driver with the PAC is to unsafely cast timer registers to a more capable type (such as `Tim4ch`), and then be very careful to use only the subset of registers that is actually supported.

Is there a way how we could both:
- provide a type-safe hierarchy of timers for users who write code generic over timers (the goal of this PR), and
- ensure that users who use the PAC to code directly against concrete peripherals don't accidentally use more channels?

CuriouslyCurious:
Hello!

I have also stumbled upon the need for reasoning about the timers on STM32's generically.

There is some documentation regarding what each family of STM32 products timer-capabilities are, described in [AN4013](https://www.st.com/resource/en/application_note/an4013-introduction-to-timers-for-stm32-mcus-stmicroelectronics.pdf). However, that is a more general overview and does not go into specifics about each board unfortunately.

honzasp:
Yes, I believe that this PR and the [accompanying PR for Embassy](https://github.com/embassy-rs/embassy/pull/3123) are at the moment the best resource about the specific differences between the STM32 timers (at least I wasn't able to find anything comparable). The taxonomy of timers that I proposed for Embassy is "optimal" in the sense that it allows you to precisely express the intersection of capabilities of an arbitrary set of timer peripherals, which means that you can write generic drivers based on timers and rest assured that you are using only the registers and register fields that are available in all timers that your driver intends to support.

Unfortunately, there is no indication that this work will ever get merged into Embassy. I created my own HAL for STM32 that solves this problem (and other problems that I have, such as removing all generics and lifetimes in drivers), but it only supports the particular chip and peripherals that I need, so it's nowhere near to being a replacement for embassy-stm32.

leftger:
The author of this PR came up with their own HAL. There's no interest in bringing this PR (afaik) into Embassy.

honzasp:
@leftger If there's some interest from Embassy, I'm happy to cooperate on merging this PR (or another PR in a similar vein), even if I have already scratched my own itch.

leftger:
Ok if you want to refactor this PR and resubmit it I'd be down to checking it out

honzasp:
This PR was meant to accompany a PR to Embassy (https://github.com/embassy-rs/embassy/pull/3123), which contains a more detailed description of the change and the rationale for it.

I haven't received any feedback on the Embassy PR; the discussion we had with Dirbaio two years ago on this PR was that to enable the Embassy PR, this stm32-data PR exposes all 4 channel registers on all timer peripherals, including those with fewer channels. I can't fix that without breaking the Embassy PR: we either give Embassy users type-safe access to the correct registers of timers in a general way, or we give PAC users the guardrail to not access nonexistent channels when writing code against a hardcoded timer peripheral.

PR #498, opened 2024-07-01
Use bool for ADC `CONT` fields

andresv:
This field enables continuous conversion mode. Using just bool for this is more appropriate.


Dirbaio (approved):
thanks!

PR #503, opened 2024-07-16
Add registers for comparators on F373

Stupremee:


data/registers/comp_f3_v1.yaml
```diff
@@ -0,0 +1,130 @@
...
+
+block/COMP2:
+  description: General purpose comparators.
+  items:
+  - name: CSR
+    description: control and status register.
+    byte_offset: 0
+    fieldset: CSR_COMP2
+
+fieldset/CSR_COMP1:
```
Dirbaio, line 17: please merge the registers for comp1 and comp2, otherwise it's too hard to add HAL support without macro abuse since they end up being different types.

if the difference is WNDWEN being available on comp2, then just make it available on both and add a comment on the description.

Stupremee: Figured as much, and had already changed that because I noticed it while writing my HAL. 
Just pushed the commit

Stupremee:
Once I'm done writing and verifying my HAL, I will mark this pull request as read :)
I just want to make sure all registers are correct and are working.

Dirbaio (approved):
thanks!

Stupremee:
Alright, everything seems to work!
After some API re-design, I will try to contribute the comparator to the embassy-stm32 crate :)

PR #507, opened 2024-07-19
Fix offsets of stm32f1 BKP registers

samueltardieu:


samueltardieu:
image

The same is true for the `RTCCR`, `CR` and `CSR` registers.

samueltardieu:
Is there anything else to do with this PR? It looks like it hasn't been reviewed yet.

samueltardieu:
As a double check and an illustration, before this patch

```rust
#![no_std]
#![no_main]

use defmt_rtt as _;
use embassy_executor::Spawner;
use embassy_stm32::pac;
use panic_probe as _;

#[embassy_executor::main]
async fn main(_spawner: Spawner) {
    defmt::info!("BKP_CR = {:#X}", pac::BKP.cr().as_ptr());
}
```

will display 

```
BKP_CR = 0x40006c2c
```

while the right address is 0x40006c30 as can be seen for example on https://github.com/stm32-rs/stm32-rs-mmaps/blob/master/stm32f103.mmap#L1930C1-L1930C65

elagil:
I verified all offsets with `RM0041`

PR #510, opened 2024-07-24
Added missing CCMR for `lptip_v1b_h7`

Dherse:
There were missing registers, enums, etc. For the LPTIM devices.


Dherse:
There is a mistake in the register definition, will fix and re-open later

Issue #509, opened 2024-07-23
Incorrect timer versions on STM32H723


elagil:
This probably also applies to other chips of the H7 variant, but it seems like the H7's timers are using the `v1` schema, when most of them should be using the `v2` schema. I noticed when trying to use the trigger source `ITR5` of `TIM2`, which is not present in `v1`.

I found the following mapping.

v1 timers:
6
7
12
13
14
15
16
17

v2 timers:
1
2
3
4
5
8
23
24

Shall I create a pull request?

elagil:
Ok! I will try to do all H7 series parts then.

PR #511, opened 2024-07-26
fix(stm32h7): timer version assignements

elagil:
Closes https://github.com/embassy-rs/stm32-data/issues/509

### DIFF
--- a/data/registers/timer_v1.yaml
+++ b/data/registers/timer_v1.yaml
@@ -1948,6 +1948,42 @@ enum/TS:
   - name: ETRF
     description: External Trigger input
     value: 7
+  - name: ITR4
+    description: Internal Trigger 4
+    value: 8
+  - name: ITR5
+    description: Internal Trigger 5
+    value: 9
+  - name: ITR6
+    description: Internal Trigger 6
+    value: 10
+  - name: ITR7
+    description: Internal Trigger 7
+    value: 11
+  - name: ITR8
+    description: Internal Trigger 8
+    value: 12
+  - name: ITR9
+    description: Internal Trigger 9
+    value: 13
+  - name: ITR10
+    description: Internal Trigger 10
+    value: 14
+  - name: ITR11
+    description: Internal Trigger 11
+    value: 15
+  - name: ITR12
+    description: Internal Trigger 12
+    value: 16
+  - name: ITR13
+    description: Internal Trigger 13
+    value: 17
+  - name: ITR14
+    description: Internal Trigger 14
+    value: 18
+  - name: ITR15
+    description: Internal Trigger 15
+    value: 19
 enum/URS:
   bit_size: 1
   variants:


elagil:
Apparently, multiple timer versions are not allowed at once. I am quite sure that the H7 parts do actually use timers of different versions.

For example, I checked the `TISEL` `byte_offset` field, which for the basic timers on the H7 indicates `timer_v1` (the offset is indeed 104, instead of 92 for `timer_v2`).

elagil:
I made a "fix" to allow mismatching timer peripheral versions in the chip generator.

stm32-metapac-gen/src/lib.rs
```diff
@@ -72,7 +72,7 @@ impl Gen {
 
             if let Some(bi) = &p.registers {
                 if let Some(old_version) = peripheral_versions.insert(bi.kind.clone(), bi.version.clone()) {
-                    if old_version != bi.version {
+                    if (old_version != bi.version) && (bi.kind != "timer") {
```
Dirbaio, line 75: this won't work when you try to actually use this with embassy-stm32.

If H7 neither matches v1 nor v2, the fix is to make a new version and make all h7 timers use that.

elagil: Ok, I understand. I will do that then.

elagil:
@Dirbaio I created a single file in the following way:
- Copy `timer_v2.yaml` to `timer_h7.yaml`.
- Remove all definitions from `timer_h7.yaml` that are the exactly same in `timer_v1.yaml`.
- Rename any remaining definitions (`enums`, `fieldsets`, `blocks`) to end in `_V2`, as well as their references.
- Copy all definitions from `timer_v1.yaml` into `timer_h7.yaml` (to sensible locations)

Is there a good way to test this?

elagil:
`embassy-stm32::timer` does not yet support multiple driver versions at once. For example, `set_trigger_source` only accepts an input of type `TriggerSource`, but there now has to be an additional `TriggerSourceV2` type for the V2 timers of the chip.

@Dirbaio Do you have a suggestion for avoiding a lot of code duplication? Of course, we can just compile for either the V1 or the V2 timers. The others would not be usable in that case (not great).

elagil:
Is this a good approach, or should we change it?

Dirbaio:
@elagil if you look through the diff the CI bot posted, there's many timers where the `registers` field is gone. This will make those be unusable with the HAL. The reason for it is no longer matching any rule in `perimap`. Can you take a look?

elagil:
@Dirbaio I know, the disabled timers are the v1 timers. This PR deliberately makes it so that only v2 timers are active.

Up to now, all timers (even v2) were used with the v1 register set. I chose to enable only v2 timers with this PR because there are more v2 than v1 timers, and they have more features. These now use the v2 register definitions, as they should.

Do you have an idea for solving this problem? v1 and v2 do not have compatible registers and cannot be merged into a single definition file. As you said, in embassy, we cannot use multiple driver versions at once.

Dirbaio:
ah sorry I should've read the thread instead of assuming my memory is in working order :sweat_smile: 

There's people out there using a lot of timers, sometimes someone asks in Matrix how to free the timer used by the time driver because they're running out. I can imagine making a bunch of timers not have registers will be unpopular. I'd say allowing all timers to be usable is better than having 100% accuracy but only allowing using half the timers.

so yes, if some timers have some features and the others don't then we'll need two copies of fieldsets/enums and make the right timer block use the right ones...

The HAL will have to deal with this ... somehow. Perhaps it's okay for now to be slightly inaccurate, for example using the "more advanced" enums/fieldsets for all timers and hope users read the docs of which features are available on which timers.

elagil:
> The HAL will have to deal with this ... somehow.

Ok, I will think of something for handling two versions at once without excessive duplication. I need some of the advanced features of the v2 timers (e.g. USB SOF signal as trigger), so I can't just use v1 definitions.

kalkyl:
Cool, just curious :)

elagil:
STM32H7 does not use v2 timers, as previously assumed.
It uses TS enum entries that were not included in the v1 definition yet.

PR #520, opened 2024-09-05
Update multiple LPTIM-related fields

Ecco:
- Interface matching of `lptim_v2a` and `lptim_v2b`
- Adding interrupts
- Updating inline documentation

### DIFF
--- a/data/registers/lptim_v2a.yaml
+++ b/data/registers/lptim_v2a.yaml
@@ -1,4 +1,4 @@
-block/LPTIM_ADV:
+block/LPTIM:
   extends: LPTIM_BASIC
   description: Low power timer with Output Compare
   items:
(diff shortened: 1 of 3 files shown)


stm32-data-gen/src/interrupts.rs
```diff
@@ -477,6 +477,7 @@ fn valid_signals(peri: &str) -> Vec<String> {
         ("I3C", &["ER", "EV", "WKUP"]),
         ("FMPI2C", &["ER", "EV"]),
         ("TIM", &["BRK", "UP", "TRG", "COM", "CC"]),
+        ("LPTIM", &["BRK", "UP", "TRG", "COM", "CC"]),
```
Dirbaio, line 480: There's no chip that splits LPTIM interrupts into several interrupt lines for the separate events (break, update, trigger, etc), so I don't think we should add this.

If you don't add it then LPTIMs get a single interrupt signal "GLOBAL", which you can still use from embassy-stm32.

Ecco: Agreed. I wasn't quite sure about this to begin with, which is why I put it in its own commit. Just nuked it 😄

PR #523, opened 2024-09-19
fix rng peripheral versions for STM32H7 series

hansihe:
The `STM32H7[23AB]` chips appear to use V2 of the RNG peripheral.

I omitted the `STM32H7[RS][37]` chips from the config as they seem to use a variation of V3.

Reference manuals to check:
* RM0433 -> V1
* RM0399 -> V1
* RM0468 -> V2
* RM0455 -> V2
* RM0477 -> Vwhat


leftger:
This has already been addressed. PR is no longer relevant.

Issue #522, opened 2024-09-17
Handling multiple simultaneous revisions of peripheral (ADC on U5)


klownfish:
I’m having trouble implementing the ADCs for the U5 series due to it having multiple simultaneous ADC revisions. Some chips have up to three ADC instances: ADC1, ADC2, and ADC4. While ADC1 and ADC2 are identical, ADC4 is a completely different revision. I haven't seen a setup like this before so I'm unsure on how to best handle it. 

The most straightforward solution in my opinion is to treat them as separate peripherals (e.g. `adc12_u5.yaml` and `adc4_u5.yaml`), but this breaks the pattern of having a generic "adc" peripheral across all chips.

An alternative would be to use a single file with separate blocks, similar to how timers are handled. However due to overlapping fields and enums everything related to ADC4 would need to be prefixed (e.g. `ADC4_xxxx`)

Are there any existing examples of this kind of situation, or is there a better way to approach it? Any guidance or feedback would be greatly appreciated! 

Thanks!

taylorh140:
I think I'm running into this same problem trying to include the FMPI2c peripherals on the stm32f446 and a few others:
 - the other I2C's on the device are "I2c_v1" type
 - But the FMPI2C is "I2c_v2" type
 
 and i get a fun error:
 
![image](https://github.com/user-attachments/assets/54edc59d-dd78-422a-8e06-e623f3347a9c)

klownfish:
Yeah this seems to be the same core problem that I ran into. There's also this issue that's similar I realized just now
https://github.com/embassy-rs/stm32-data/issues/509

PR #525, opened 2024-09-25
Add support for all U5 ADCs

klownfish:
This merge request adds register definitions for all ADCs on the U5 series. Currently only ADC4 is supported.
 
The reason for this is that not all ADCs on the U5 are of the same version. Some are similar to V4 (ADC1/2), while others are closer to V3 (ADC4). To work around this I have prefixed everything related to ADC4 and instantiated both versions as blocks in `adc_u5.yaml`. This approach seemed to produce the cleanest results when implementing the driver in embassy.

I refer to ADC4 explicitly everywhere which might seem questionable but I believe it’s the best approach. It has been a constant pattern throughout the U5 series and given how the modules are numbered (1, 2, and 4) I think it's clear that 4 is the outlier.

First draft of the embassy driver:
https://github.com/klownfish/embassy/tree/u5_adc/embassy-stm32/src/adc

related issue:
https://github.com/embassy-rs/stm32-data/issues/522


MDr164:
Looks good to me. I'll give it a try in conjunction with your updated ADC driver for the U5 series this week.

klownfish:
Awesome! I cleaned it up a bit and made a pull request

PR #526, opened 2024-09-26
Adding FMPI2C to F4xx family.

taylorh140:
This adds the FMPI2C i2c devices, but it was a bit difficult because only one i2c version is allowed per chip. This was fixed by adding a symlink to i2c_v2.yaml which was named i2c2_v2.yaml allows both devices to exist on the same chip and keeps the yaml file contents aligned.


Dirbaio:
i'm a bit worried about the symlink. Does it work well if you clone the repo from Windows and run the generation?

taylorh140:
Good question. I can try it later. Or if you prefer just replace it with a copy. Ntfs supports symbolic links afaik.

taylorh140:
So @Dirbaio  I would test that, but it looks like windows builds are broken on main:
image

So, I as long as they don't work... I can't really tell if they stop working with my changes.   🤣 windows, right?

taylorh140:
So I figured out I could use WSL to do the d download-all in the folder and then execute the rest on windows. And your right! something doesn't work well:
image
So I'll go ahead and make the change.

Dirbaio:
thanks!

PR #533, opened 2024-10-31
Add const FLASH_ERASE_VALUE

ExplodingWaffle:
Adds a const for erase value next to the other flash consts. (working on adding erase value to embedded-storage and embassy-boot. this seems nice to have anyway)


Dirbaio:
This info is already exposed in `METADATA`. `embassy-stm32` can read it from there in `build.rs` and generate consts or use it in `embedded-storage` impls. 

I'd prefer avoiding adding more consts `stm32-metapac`, and gradually change everything to go through `METADATA`. It's more "general" than consts, it allows consumers to generate anything they want in build.rs. (The existing consts predate `METADATA`, they're from a time `stm32-metapac` exported info through very cursed macros instead)

Issue #519, opened 2024-09-05
Missing register definitions for configuring USB OTG HS PHY (only for STM32U59x/5Ax/5Fx/5Gx)


wagcampbell:
I'm running into some issues getting USB OTG working for an STM32U59x part.

Unlike the other STM32U5 parts, there are a some additional steps required to configure this USB OTG HS for STM32U59x/5Ax/5Fx/5Gx.
![image](https://github.com/user-attachments/assets/1dfd7771-c627-4c95-b021-f4eefd10e855)
![image](https://github.com/user-attachments/assets/30a33eb9-471a-4143-8359-d3558d6d5d5c)
![image](https://github.com/user-attachments/assets/cd214887-6d4d-41e5-a84b-c1f0a948b189)

The stm32 hal is missing some register definitions:
![image](https://github.com/user-attachments/assets/822fea77-7e42-4b5e-b757-0e152bfebec1)
![image](https://github.com/user-attachments/assets/7387a82f-e9c5-41a9-a388-321c09289e56)
![image](https://github.com/user-attachments/assets/16a15b04-0f8f-4f82-8d1f-98b5cadb1ba3)

Also, while the OTG_FS and OTG_HS control and status registers (CSRs) are mostly the same, there are some differences.
Based on what I've gathered, I expect this will mean a change to the embassy-usb-synopsys-otg crate.

I haven't been able to successfully get this working yet, so there may be some additional steps/changes.

Hopefully, I can manage to get this working and can follow up with more details.

MDr164:
I added the missing register definitions and added the extra clock to rcc. Last two bits are the setup sequence for these chips and potentially an addition to the synopsys otg crate. I'll try getting to that and testing on hardware (I got an STM32U5A5 for that) rather sooner then later.

PR #538, opened 2024-11-11
Add missing STM32U5 SYSCFG and PWR registers

MDr164:
This adds some missing register definitions as noted in #519 to enable the USB OTG_HS PHY on some STM32U5 platforms.

### DIFF
--- a/data/registers/pwr_u5.yaml
+++ b/data/registers/pwr_u5.yaml
@@ -467,6 +467,10 @@ fieldset/UCPDR:
 fieldset/VOSR:
   description: voltage scaling register
   fields:
+  - name: USBBOOSTRDY
+    description: "OTG_HS EPOD booster ready\r This bit is set to one by hardware when the power booster startup time is reached. The OTG_HS clock can be provided only after this bit is set."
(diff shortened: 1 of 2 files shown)


Dirbaio (approved):

Dirbaio:
lgtm, any reason it's draft?

MDr164:
Those registers are not available on all STM32U5 chips, only the ones with an internal USB OTG_HS PHY. Not sure if there needs to be some check somewhere?

MDr164:
> lgtm, any reason it's draft?

I'm not sure if I need to implement handling of this note from the datasheet here:

> This bit is only available on some devices in the STM32U5 Series. Refer to the device
datasheet for availability of its associated peripheral. If not present, consider this bit
reserved and keep it at reset value.

Dirbaio:
no need! we usually "merge" all the syscfg, rcc registers from a family into a single version, otherwise there'd be way too many versions. The user/hal is trusted to not access those on chips where they're not present.

(Only need to split is when they're actually incompatible, for example bit X is at one address in chip A and in another in chip B)

PR #546, opened 2024-12-10
STM32U5: Add HSPI support

williams-one:
I tried to add the support for HSPI peripheral on STM32 U5
* The peripheral is present only on some chips of the U5 family.  I got nothing for the other families when running `./d extract-all HSPI1`.  And there was basically only one version of it 
* I have confirmed the value of the GPDMA channel in the RM for U5


williams-one:
I have added some additional info for reference.
I'm still learning, so please let me know if something is wrong and has to be done differently

* Families of STM chips with HSPI support (only U5) and devices with HSPI support within the U5 family
image

* GPDMA channel
image

* The address of HSPI register is changed in the generated data, since it was picking up the wrong value from the headers:  `0xA0000000` = `2684354560` is the memory address where the data are memory mapped.
This has been fixed using the right define in `headers.rs` and now the proper address is used `0x420D3400` = `1108161536`

image

image

Dirbaio (approved):
thanks!

PR #547, opened 2024-12-13
More HRTIM updates

dlaw:
Sorry for the trickle of separate HRTIM PRs.... I keep thinking I'm done and then I find something else.

In this PR:
1. Combine SDTR and SDTF enums, since they're the same. (possible values are "positive" or "negative")
2. Combine TAU/TBU/TCU/TDU/TEU into a length-5 array.


dlaw:
2024-12-12-212908_1265x603_scrot
2024-12-12-212933_1281x3046_scrot

PR #548, opened 2024-12-15
Extract DMA remapping data for F0/F3

fwolter:
This is the basis to fix embassy-rs/embassy#3643.


Dirbaio:
Putting a single bool in the generated data is too "ad hoc". I think the data should contain the full info of which peripheral+register+field to set to enable the remap. 

Reasons:

- The bool is specifying the value, but it's not specifying *which field*. This makes users (like the embassy-stm32 PR you sent) to search it by name themselves. It's nicer if the data is "as cooked as possible" so using it is easier.
- There's more chips with "remaps" like this, for example F1 for GPIO. These are on different peripherals (AFIO) and registers (MAPR, MAPR2). The data model should ideally be the same for all chips, so remaps can be handled the same way.

For example it could be this:

```json
{
  "signal": "ADC2",
  "channel": "DMA2_CH1",
  "remap": {
    "peripheral": "SYSCFG",
    "register": "CFGR1",
    "field": "TIM17_DMA_RMP",
    "value": 0
  }
}
```

fwolter:
Thank you very much for your feedback! I will incorporate it. I wasn't aware there were even more remapping bits.

fwolter:
I agree, that'd be much easier! How about introducing a new category in the family extra YAML files to define the mappings? Would work for the F0 family - the remaps are all the same within the family, but within F3 there are at least three different maps. Or introduce `data/dma/` with dedicated YAML files?

Dirbaio:
either a new yaml, or just put the data in a .rs file like we already do with the memory maps.

fwolter:
That took longer than expected to get all the device types and their IP versions straight. The ridiculous number of copy-paste errors and typos in the reference manuals didn’t help, though 😅 But in the end, it turned out relatively simple.

I also implemented your review feedback in the PR over in embassy.

Dirbaio:
looking good! two things:

- There's remaps that are split in two bits. I think it'd be cleaner to merge them into a single 2-bit field, then represent the values as numbers 0,1,2,3. This'd allow changing `remap` to be one single object, not an array.

```json
{
  "signal": "CH1",
  "channel": "DMA1_CH4",
  "remap": [
    {
      "peripheral": "SYSCFG",
      "register": "CFGR1",
      "field": "TIM16_DMA_RMP",
      "value": "true"
    },
    {
      "peripheral": "SYSCFG",
      "register": "CFGR1",
      "field": "TIM16_DMA_RMP2",
      "value": "false"
    }
  ]
},
```

- Making the values `"true"` and `"false"` as strings is a bit strange. I'd put `0`,`1` instead and make embassy-stm32 deal with this. It'll be more consistent with the 2-bit fields like the example above. Also, the JSON data is supposed to be language-agnostic, the fact that chiptool chooses to generate 1bit fields as bools is rust-specific.

Dirbaio:
> - *_RMP and _*RMP2 are distinct fields in the register, they are not adjacent bits. Technically, it'd make sense to set them in one instruction, but then we'd need to include a bitmask (some also need to be cleared) and drop the field name.
> - STM32F334x has bits split across CFGR1 and CFGR3 for a single remap.

WTF.

fwolter:
Thank you!

PR #550, opened 2024-12-20
Fix BF1 enum

CrystalKeck27:
See [stm32f75xxx and stm32f74xxx reference manual](https://www.st.com/resource/en/reference_manual/dm00124865-stm32f75xxx-and-stm32f74xxx-advanced-arm-based-32-bit-mcus-stmicroelectronics.pdf) revision 8, section 18.7.21

I do not know with certainty that this change applies to all micros, because, to be quite frankly honest, I don't want to read 30 datasheets. However, I checked the one linked above and two others from different families (u5 and h7) and this change tracks in all of them.


Dirbaio:
thanks!

Issue #531, opened 2024-10-22
STM32G4 dual flash bank configuration


u6bkep:
I was having some trouble with the flash erase routines in the STM32g4 HAL, so I did some digging. The first issue I cam across was missing register definitions, so I fixed those. [pull request incoming](https://github.com/u6bkep/stm32-data/commit/3d3605c9f9e7239e4629f2d58523334275294cac). Then I noticed the G4, category 3 devices at least, default to dual bank mode meaning any attempt to erase any data in the second half of ROM will fail because the page offset is calculated incorrectly. I have confirmed erasing a page in bank 2 can be done by running the erase sequence with manual register writes.
![rm0440 v8](https://github.com/user-attachments/assets/6bbd4ec0-6b0d-4431-8b4a-ce7a4dfc6627)

now I need some advice on how to move forward with this chip. Is there a recommended way in the HAL to configure single bank mode so the erase function works? or is the correct solution to modify the the [metadata](https://github.com/embassy-rs/stm32-data-generated/blob/400b02bed6679118b8b78bd219ba00273fae5bf8/stm32-metapac/src/chips/stm32g474re/metadata.rs) to reflect the two memory banks?

Dirbaio:
The metadata and the embassy-stm32 implementation are for single-bank mode.

The problem is the firmware itself can't switch the chip from dual to single bank mode since it shuffles the flash, corrupting the firmware. You have to set it beforehand before flashing. I think ST's tools can do it, maybe probe-rs can also do it by writing the right value by flashing to that option byte address. 

I think the only "fix" we can do in embassy-stm32 is to detect the chip is in dual-bank mode and panic, with a message for the user.

Dual-bank G4s are a bit cursed because there's a *hole* between the banks (unless you have the biggest flash size, 512kb). So you can't pretend they're a single bigger contiguous flash region. Erase size also changes: 2kb in dual bank, 4kb in single bank. So i'd recommend sticking to single bank if you can.

zhihu2andy:
I am wondering if the issue had been fixed. I still experienced the same issue on STM32G473CB chip with 128 KB flash and there is a 384 KB hole between 2 banks. I want to erase the last 2 pages (2*2048 KB) in the the flash.
Erase last 2 pages fails with dbg message print out.
 0.000122 [TRACE] Erasing from 0x804f000 to 0x8050000 (<crate> └─ <invalid location: defmt frame-index: 14>:0)
[...]

PR #551, opened 2024-12-20
add missing registers to stm32g4 category 3 flash peripheral

u6bkep:
I went through the reference manual carefully to find discrepancies in `flash_g4c3.yaml` and added anything missing and fixed a few errors. Importantly now a program can check the user config words to check for dual bank mode even if the only thing it can do about it is print an error message.


u6bkep:
#531

Dirbaio:
thanks!

PR #556, opened 2025-01-02
update STM32G0 flash register access methods

chanterheld:
Implements a union of flash functionality described in [RM0454 Rev 5](https://www.st.com/resource/en/reference_manual/rm0454-stm32g0x0-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) and [RM0444 Rev 6](https://www.st.com/content/ccc/resource/technical/document/reference_manual/group0/2f/21/cb/33/78/80/42/64/DM00371828/files/DM00371828.pdf/jcr:content/translations/en.DM00371828.pdf).
Changes are to support chips with larger flash capacity and dual memory banks.

 ### Flash control register
Added BKER and MER2 bits to  to support page and mass erasing of bank 2
Extended Page number access bit size to 10 bits to support bigger flash single bank devices (currently causes a wraparound beyond page 63)

 ### Flash option register
Added bank swap and dual bank operation enable bits. 

### PCROP (A/B 1/2 start/end) and WRPxxR
Extended the width to support more flash pages

 ### Flash status register
Added bank 2 busy bit

### Bank 2 support
Added the following registers for bank 2
- ECCR2
- PCROP2ASR
- PCROP2AER
- WRP2AR
- WRP2BR
- WRP2BR
- PCROP2BER


Dirbaio:
could you split it instead in two versions `flash_g0x0` and `flash_g0x1`? this way we don't expose registers/fields that don't exist in g0x0. thanks!

Dirbaio:
thank you! :pray:

PR #557, opened 2025-01-02
Update to chiptool for more Debug impls and defmt::Format impls.

de-vri-es:
Update to the latest `chiptool` version, to add `Debug` and `defmt::Format` impls for fieldsets, value enums and interrupt enums :partying_face: 

Follow up for https://github.com/embassy-rs/chiptool/pull/44

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,56 +4,57 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.2"
+version = "1.1.3"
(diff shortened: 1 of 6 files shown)


de-vri-es:
Hmm, empty diff? That seems weird.. Locally I do get the `defmt::Format` impls.

Dirbaio:
> I guess the diff created in CI is somehow incorrect?

the diff is only from the json files. (it's there so it's easy to review the impact of changes in Rust code extracting the info. If that's OK I just assume the rust PAC is OK as well)

PR #559, opened 2025-01-04
Map h7rs TMPSENS* RCC fields to the DTS peripheral

yodaldevoid:
Despite the reference manual and header files referring to these fields in terms of DTS, the SVD files insist on using TMPSENS.

I opted to do the mapping at lookup time rather than renaming all of the fields.


Dirbaio:
thanks!

PR #563, opened 2025-01-20
Removes double offsets from bits

taylorh140:
As far as I can tell these are just noise in the file. They also incorrectly affect the bit_size field. The first instance seems to be correct and the rest are just bad as far as i can tell.


taylorh140:
image
CHMOD is two bits. So ill have to figure out what is going on here.

PR #564, opened 2025-02-01
Add missing ALRWF bit to ICSR register in RTC v3 yamls

markus-k:
I'm currently working on implementing RTC alarms in `embassy-stm32` and noticed the `ALRWF` bit in the `ICSR` was missing in `stm32-metapac` for v3. So this PR adds the definition to the the RTC v3 definitions.


Dirbaio (approved):
thanks!

PR #565, opened 2025-02-02
Fix naming of ALRIE flag in rtc_v3

markus-k:
All other variants name this flag ALRIE, but was named ALRAIE for v3. Probably a copy/paste error or missed rename.


Dirbaio (approved):
thanks!

PR #567, opened 2025-02-13
Add OTP flash memory configuration

rmja:
This PR restores the OTP configuration removed in https://github.com/embassy-rs/stm32-data/pull/460.
I hope that I got the configuration right for all the series. The regions vary in size from 528 bytes (yes 16*32+16) to 512 bytes and 1k and 2k with many different locations. The data is based on the information in the header files and extracted with a `grep OTP *.h`.

Note that I had to make a new `mem` macro variant because of the odd sizes. Don't know if that is the best approach.

I have verified that it is correct for stm32f429 that I am on myself.

### DIFF
--- a/stm32-data-gen/src/memory.rs
+++ b/stm32-data-gen/src/memory.rs
@@ -31,19 +31,27 @@ const fn access(access: &'static str) -> Access {
 }
 
 macro_rules! mem {
-    (@row $name:ident $addr:literal $size:literal) => {
+    (@row $name:ident $addr:literal $size_kb:literal) => {
         Mem {
             name: stringify!($name),
             address: $addr,
-            size: $size*1024,
+            size: $size_kb*1024,
             access: None,
         }
     };
-    (@row $name:ident $addr:literal $size:literal $access:ident) => {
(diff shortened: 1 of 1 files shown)


rmja:
I have updated so the flash settings now come with an `erase_size: 0`.
It should now be ready for review.

Dirbaio (approved):
thanks!

ninjasource:
Hi @rmja,

Is there something I need to configure in Embassy to get it to compile with the latest stm32-metapac? I am using an `stm32u5g9zj` and get the following compile error:

```
error[E0412]: cannot find type `OTPRegion` in module `crate::_generated::flash_regions`
```

which happens here: `embassy/embassy-stm32/src/flash/common.rs:247:1`

If this is a known thing and all I need to do is to apply your embassy [PR](https://github.com/embassy-rs/embassy/pull/3881) to my branch then never-mind, just checking.

rmja:
@ninjasource  I have just verified, and https://github.com/embassy-rs/embassy/pull/3881 will fix the issue that you see. Lets hope that it will get merged soon.

PR #571, opened 2025-02-27
Added STM32L4 OPAMP data

hvraven:
I currently work on an L4 where I need the OpAmp and had to find out the definitions are missing. This is my attempt at extracting the register information following the documentation in the Readme. I extended the existing transformers from other Opamps and used the extracted `l4x1.yaml` as base. All the `l4xx.yaml` are identical, except a description error in the `l4x3.yaml`. The other l4-Variants (`l412`, `l4[pr]x`) appeared to be diffenent.

I do have some questions:
* Some larger variants have two opamps. The existing transformers delete those. The register map for both OpAmps is identical with a different offset. So I think this is fine, but not really sure how this is handled.
* I applied some manual changes to the yaml besides the transformers:
  * For the descriptions I replaced with 's/OPAMP1/OPAMP/'
  * I had to manually delete the `OPAMP2` references in the `block/OPAMP`.

### DIFF
--- a/data/registers/opamp_l4.yaml
+++ b/data/registers/opamp_l4.yaml
@@ -1,0 +1,176 @@
+block/OPAMP:
+  description: Operational amplifiers.
+  items:
+  - name: CSR
+    description: OPAMP control/status register.
+    byte_offset: 0
+    fieldset: CSR
+  - name: OTR
+    description: OPAMP offset trimming register in normal mode.
+    byte_offset: 4
+    fieldset: OTR
+  - name: LPOTR
+    description: OPAMP offset trimming register in low-power mode.
+    byte_offset: 8
+    fieldset: LPOTR
+fieldset/CSR:
+  description: OPAMP control/status register.
+  fields:
+  - name: OPAEN
+    description: Operational amplifier Enable.
+    bit_offset: 0
+    bit_size: 1
+  - name: OPALPM
+    description: Operational amplifier Low Power Mode.
+    bit_offset: 1
+    bit_size: 1
+    enum: OPALPM
+  - name: OPAMODE
+    description: Operational amplifier PGA mode.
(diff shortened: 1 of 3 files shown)


Dirbaio:
> Some larger variants have two opamps. The existing transformers delete those. The register map for both OpAmps is identical with a different offset. So I think this is fine, but not really sure how this is handled.

what you did looks correct: you add the entry to perimap and both peripherals get matched to the register block:

```diff

diff --git a/data/chips/STM32L4A6VG.json b/data/chips/STM32L4A6VG.json
index 62fa668..f535499 100644
--- a/data/chips/STM32L4A6VG.json
+++ b/data/chips/STM32L4A6VG.json
@@ -5577,6 +5577,11 @@
         {
           "name": "OPAMP1",
           "address": 1073772544,
+          "registers": {
+            "kind": "opamp",
+            "version": "l4",
+            "block": "OPAMP"
+          },
           "rcc": {
             "bus_clock": "PCLK1",
             "kernel_clock": "PCLK1",
@@ -5607,6 +5612,11 @@
         {
           "name": "OPAMP2",
           "address": 1073772560,
+          "registers": {
+            "kind": "opamp",
+            "version": "l4",
+            "block": "OPAMP"
+          },
           "rcc": {
             "bus_clock": "PCLK1",
             "kernel_clock": "PCLK1",
```

hvraven:
Thanks for the quick response: One other thing I ran into: The L4 calls the enable register `opaen`, while in the other variants `opampen` is used. Should I change the name in the definition, even though it does not match the description in the reference manual. Otherwise conditional checks are required in the HAL to change simply to change the register name.

Dirbaio:
yeah, we should change it. We give priority to consistency between families over consistency with the RMs.

PR #577, opened 2025-03-10
Switch stm32u5 to dma2d-v2 and manually corrected yaml

ninjasource:
Hi,

I have been busy building out the DMA2D driver for embassy-stm32 and I noticed some small issues relating to auto-generated config. The DMA2D peripheral is used for graphics applications and has the potential to vastly reduce render times on large displays. I used section 20 of rm0456 to double check everything. The changes are outlined as follows:

- Changed the documentation to make grammatical sense (cosmetic but nice).
- Fixed the `OCOLR` which was incorrectly represented due to an unusual setup in the reference manual (the register is duplicated to demonstrate all possible permutations of color modes). IMO. it is better to expose the raw register value (u32) and handle the permutations in code since as all bit values are valid but mean different things depending on setup in another register.
- Added `LOM` to `CR` register - Line offset mode
- Added 2 more operating modes: MemoryToMemoryPFCBlendingFixedColorFG and MemoryToMemoryPFCBlendingFixedColorBG that were missing. These can be gated in code for mcu's that don't support them but they would simply be reserved on those chips anyway.
- Make the perimap use the `dma2d_v2.yaml` for `stm32u5` chips which, being very recently released chips, I am fairly certain support v2.

### DIFF
--- a/data/registers/dma2d_v2.yaml
+++ b/data/registers/dma2d_v2.yaml
@@ -6,7 +6,7 @@ block/DMA2D:
     byte_offset: 0
     fieldset: CR
   - name: ISR
-    description: DMA2D Interrupt Status Register
+    description: DMA2D interrupt status register
     byte_offset: 4
     access: Read
     fieldset: ISR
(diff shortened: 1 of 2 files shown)


Dirbaio:
very thanks!

PR #590, opened 2025-04-18
feat: add stop mode rules which override the fallback behavior

ckrenslehner:
Allows adding custom rules for the lowest possible stop mode for a peripheral.
This affects the embassy low-power executer, as the peripheral limits the available power modes when it's active.

Supporting information for the rules below:

I went to the `sources/headers` directory and ran `rg` for my MCU.
{8083D2CE-04CF-4157-8866-F8064F773234}

Checking the manual of the stm32wb55 shows the following:

{B8E76055-DD32-4DEE-AEE8-B7C87D2CD4F4}

{9B04AC08-117B-42DA-91A5-7B9E9E9C87C6}

So the header file is not accurate enough to differentiate between stop1 and stop2, but it's an indication at least.
According to the [manual](https://www.st.com/resource/en/reference_manual/rm0434-multiprotocol-wireless-32bit-mcu-armbased-cortexm4-with-fpu-bluetooth-lowenergy-and-802154-radio-solution-stmicroelectronics.pdf), the USART and the I2C should just work in lower power modes as well. At least as far as I understand.

### DIFF
--- a/stm32-data-gen/src/low_power.rs
+++ b/stm32-data-gen/src/low_power.rs
@@ -1,0 +1,59 @@
+use stm32_data_serde::chip::core::peripheral::rcc::StopMode;
+
+use crate::util::RegexMap;
+
+/// Get the stop mode limit for a peripheral based on the MCU and peripheral name.
+/// Determines the lowest possible stop mode when a peripheral is enabled.
+///
+/// Parameters:
+/// - mcu_name: the full name of the MCU (e.g., "STM32WB55RG")
+/// - peripheral: the name of the peripheral (e.g., "USART1")
+pub(crate) fn peripheral_stop_mode_info(mcu_name: &str, peripheral: &str) -> Option<StopMode> {
+    /// Regexmap where the key is mcu_name:peripheral and the value is the stop mode.
+    /// Example: STM32WB55RG:USART1 -> StopMode::Stop2
+    #[rustfmt::skip]
+    static STOP_MODE_OVERRIDE_RULES: RegexMap<StopMode> = RegexMap::new(&[
+        (r"^STM32WB55.*:LPTIM1", StopMode::Standby),
+        (r"^STM32WB55.*:USART1", StopMode::Stop2),
+        (r"^STM32WB55.*:LPUART1", StopMode::Standby),
+        (r"^STM32WB55.*:I2C1", StopMode::Stop2),
+        (r"^STM32WB55.*:I2C3", StopMode::Standby),
+        
+
+        // __ATTENTION__: Keep these rules at the bottom to grant precedence to the more specific rules above
+        // Every peripheral with LP prefix is assumed to be able enter up to Stop1 mode
+        (r".*:LP.*", StopMode::Stop2), 
+        // The RTC peripheral is assumed to be able to enter up to Stop2 mode
+        (r".*:RTC", StopMode::Standby),
+    ]);
+
+    STOP_MODE_OVERRIDE_RULES
+        .get(&format!("{mcu_name}:{peripheral}"))
+        .cloned()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_get_peripheral_stop_mode_info() {
+        // MCU independent rule for RTC
+        assert_eq!(peripheral_stop_mode_info("my-test-mcu", "RTC"), Some(StopMode::Standby));
+
+        // No rule for this but starting with LP prefix, so assumed to be Stop2
+        assert_eq!(
+            peripheral_stop_mode_info("my-test-mcu", "LPTIM2"),
+            Some(StopMode::Stop2)
+        );
+
+        // MCU independent rule for RTC. Must match RTC exactly
+        assert_eq!(peripheral_stop_mode_info("my-test-mcu", "RTC1"), None);
+
+        // Rule covering all STM32WB55 for LPTIM1
+        assert_eq!(
+            peripheral_stop_mode_info("STM32WB55RG", "LPTIM1"),
+            Some(StopMode::Standby)
+        );
+    }
+}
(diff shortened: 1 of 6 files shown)


ckrenslehner:
Unfortunately, I could not find any related information in the db files in the `sources` dir. 
Maybe the rules can be encoded in the yaml files directly.

ckrenslehner:
> Unfortunately, I could not find any related information in the db files in the `sources` dir. Maybe the rules can be encoded in the yaml files directly.

I just checked the [manual](https://www.st.com/resource/en/reference_manual/rm0478-multiprotocol-wireless-32bit-mcu-armbased-cortexm4-with-fpu-bluetooth-low-energy-54-radio-solution-stmicroelectronics.pdf) of the stm32wb10cc.
The sleep capabilities with the USART and I2C are consistent.
However, this chip does only have one UART and one I2C.

Probably it would be better to check for e.g. `wb55` specifically instead of `wb`. However, I do not know if we want to split down the yaml files further. Maybe a better solution would be a `stopmode` part in the yaml where the rules can be specified for a distinct MCU and peripheral.

So e.g. in the `wb.yaml` there could be something like this:
``` yaml
stop/wb55:
  rules:
  - peripheral: LPUART1
    stopMode: Standby
  - peripheral: USART1
    stopMode: Stop2  
  - peripheral: I2C1
    stopMode: Stop2
  - peripheral: I2C3
    stopMode: Standby
```

Dirbaio:
Thanks for the PR!

- Could you keep it in rcc.rs (or make a new file?). chips.rs is already big enough.
- if you make it a regexmap where the key is `chip_name:peri_name` (for example `STM32WB55RG:LPUART1` then it's possible to write the old rules with regexes (for example `".*:LPUART\d+" -> Stop2`). This way you don't need `if`s for the old rules plus a regexmap to override, you can just have a regexmap and that's it.

ckrenslehner:
Thanks for your feedback and for the nice idea of adding the old rules into the same regex.
Now this should to the trick :-)

stm32-data-gen/src/rcc.rs
```diff
@@ -250,7 +245,8 @@ impl ParsedRccs {
                             },
                             reset,
                             bus_clock: clock,
-                            stop_mode,
+                            // The default mode is overridden in `low_power.rs`
+                            stop_mode: StopMode::default(),
```
Dirbaio, line 249: instead of leaving it at default here and editing it later, you could do the regexmap lookup here.

(if you need the chip name, just pass it in)

stm32-data-gen/src/low_power.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use crate::util::RegexMap;
+
+/// Get the stop mode limit for a peripheral based on the MCU and peripheral name.
+/// Determines the lowest possible stop mode when a peripheral is enabled.
+///
+/// Parameters:
+/// - mcu_name: the full name of the MCU (e.g., "STM32WB55RG")
+/// - peripheral: the name of the peripheral (e.g., "USART1")
+pub(crate) fn get_peripheral_stop_mode_override(mcu_name: &str, peripheral: &str) -> Option<StopMode> {
```
Dirbaio, line 11: this is not an "override" anymore if it's the only source of the info.

stm32-data-gen/src/rcc.rs
ckrenslehner: Do I have the full chip name here?
I moved it outside, because it was just wb instead of wb55.
Maybe I misunderstood but I think the whole name of the chip is not there yet when parsing the registers

stm32-data-gen/src/low_power.rs
ckrenslehner: Will change it to `peripheral_stop_mode_info`

stm32-data-gen/src/rcc.rs
Dirbaio: oh right this runs *before*. okay let's keep it this way then.

stm32-data-gen/src/low_power.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// Regexmap where the key is mcu_name:peripheral and the value is the stop mode.
+    /// Example: STM32WB55RG:USART1 -> StopMode::Stop2
+    #[rustfmt::skip]
+    static STOP_MODE_OVERRIDE_RULES: RegexMap<StopMode> = RegexMap::new(&[
+        (r"^STM32WB55.*:LPTIM1", StopMode::Standby),
+        (r"^STM32WB55.*:USART1", StopMode::Stop2),
+        (r"^STM32WB55.*:LPUART1", StopMode::Standby),
```
Dirbaio, line 18: Are you sure this is correct? I don't think LPUARTs can work in standby.

ckrenslehner: The can work in STOP2.
The StopMode enum limits it up to the certain level.
That's the enum. I just changed the comments to doc comments
``` rust
  pub enum StopMode {
      #[default]
      /// Peripheral prevents chip from entering Stop1
      Stop1,
      /// Peripheral prevents chip from entering Stop2
      Stop2,
      /// Peripheral does not prevent chip from entering Stop
      Standby,
  }
```

ckrenslehner: But maybe we could adjust the naming of the enum to be more intuitive

Dirbaio: oh wow yes this is somewhat unintuitive. Intuitively i'd say "Standby" means "works in Standby", not "works in all modes up to but not including Standby"

Dirbaio (approved):
thanks!

stm32-data-gen/src/low_power.rs
ckrenslehner: Yeah. But I am not sure if the name can be changed easily because of the metapac gen and so on

PR #593, opened 2025-04-22
Added H7RS ADC

alvidrezluke:
I have been working with the H7S3L8 on the Nucleo for a little bit now and had made these changes locally, but finally had the time to create a PR. Please let me know if I need to update the documentation strings. The ADC was tested with a H7S3L8 Nucleo and 7 sampling channels and appears to work correctly.


Dirbaio:
looking great, thank you ! :)

PR #596, opened 2025-04-24
Add configurable bank support

diondokter:
Support is there for all series, but I've only implemented it for the G4 for now


diondokter:
Ok, added support for the other chips with configurable dual bank too.
I've tried to be careful, but it's really easy to not see a small detail.

There's a question around the L5 series too.
The reference manual says it's supported in general, but then only the 512kb chips get their memory layout specified... So only support for 512kb chips?

That's what I've done now at least...

lulf (approved):

PR #602, opened 2025-05-01
Add RegexSet

bugadani:
No doubt this is a no-go :D

### DIFF
--- a/stm32-data-gen/src/util.rs
+++ b/stm32-data-gen/src/util.rs
@@ -56,3 +56,19 @@ impl<'a, T> RegexMap<'a, T> {
         res
     }
 }
+
+pub struct RegexSet<'a> {
+    map: RegexMap<'a, ()>,
+}
+
+impl<'a> RegexSet<'a> {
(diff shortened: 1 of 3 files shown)


stm32-data-gen/src/util.rs
```diff
@@ -56,3 +56,19 @@ impl<'a, T> RegexMap<'a, T> {
...
 }
+
+pub struct RegexSet<'a> {
+    map: RegexMap<'a, ()>,
+}
+
+impl<'a> RegexSet<'a> {
+    pub const fn new(map: &'a [&'a str]) -> Self {
+        Self {
+            map: RegexMap::new(unsafe { std::mem::transmute(map) }),
```
Dirbaio, line 67: lol

Dirbaio:
thanks!

stm32-data-gen/src/util.rs
bugadani: wait this was merge-worthy? 🤣

Dirbaio: yolo

PR #603, opened 2025-05-02
Add support for STM32C0[579] chips

charliefoxtwo:
Merged yaml for STM32C07 chips. Running `cargo build --features stm32c071rb` in embassy-stm32 works with this.

EDIT: removed error: issue was that I was mixing local and crates.io embassy crates.

[Matrix thread](https://matrix.to/#/!YoLPkieCYHGzdjUhOK:matrix.org/$GOtlq_F-u_he1Qa1-lVQP3OMq8CfVY_AfV30LFZRE4g)


data/registers/rcc_c0.yaml
```diff
@@ -1,931 +1,2386 @@
...
+    name: B_0x0
+    value: 0
+  - description: Interrupt pending.
+    name: B_0x1
+    value: 1
+enum/HPRE:
+  bit_size: 4
+  variants:
+  - description: '2.'
+    name: B_0x8
```
charliefoxtwo, line 214: The format of this line appears to be one of the issues resulting in the error in `parse_num` - I guess this is normally like `Mul2` or something?

data/registers/rcc_c0.yaml
```diff
@@ -13,17 +13,24 @@ block/RCC:
...
+    access: Read
+    fieldset: CRRCR
   - name: CIER
     description: RCC clock interrupt enable register
     byte_offset: 24
     fieldset: CIER
   - name: CIFR
     description: RCC clock interrupt flag register
     byte_offset: 28
+    access: Read
```
charliefoxtwo, line 28: Unsure if this should be added

data/registers/rcc_c0.yaml
```diff
@@ -13,17 +13,24 @@ block/RCC:
...
     fieldset: CIFR
   - name: CICR
     description: RCC clock interrupt clear register
     byte_offset: 32
+    access: Write
```
charliefoxtwo, line 33: Unsure if this should be added

data/registers/rcc_c0.yaml
```diff
@@ -73,10 +80,14 @@ block/RCC:
     description: RCC APB peripheral clock enable in Sleep/Stop mode register 2
     byte_offset: 80
     fieldset: APBSMENR2
-  - name: CCIPR
-    description: RCC peripherals independent clock configuration register
+  - name: CCIPR1
+    description: RCC peripherals independent clock configuration register 1
     byte_offset: 84
-    fieldset: CCIPR
+    fieldset: CCIPR1
```
charliefoxtwo, line 86: Is the preferred course of action here to leave the name as is, or rename it? Renaming will require a one-line change here:
https://github.com/embassy-rs/embassy/blob/7a97825427b58934329e0ed015d1c201db09251c/embassy-stm32/src/rcc/c0.rs#L190

data/registers/rcc_c0.yaml
charliefoxtwo, line 1: Interestingly, the bit_size for this line changed from 2 -> 1 when generating from the c071 file. Unclear why, but I left it unchanged in this PR.
https://github.com/embassy-rs/stm32-data/blob/94978672d3d1f66c5032becb17d19a8a765d1383/data/registers/rcc_c0.yaml#L525

data/registers/rcc_c0.yaml
```diff
@@ -324,7 +395,7 @@ fieldset/APBSMENR2:
     description: "ADC clock enable during Sleep mode\r Set and cleared by software."
     bit_offset: 20
     bit_size: 1
-fieldset/CCIPR:
+fieldset/CCIPR1:
```
charliefoxtwo, line 398: Alongside the earlier question, wondering if this should be changed

stm32-data-gen/src/rcc.rs
```diff
@@ -341,6 +341,8 @@ impl ParsedRccs {
                         maybe_kernel_clock = "PLL1_VCO_DIV_2".to_string();
                     } else if rcc_version.starts_with("h7rs") {
                         maybe_kernel_clock = "USB".to_string();
+                    } else if rcc_version.starts_with("c0") {
+                        maybe_kernel_clock = "USB".to_string();
```
charliefoxtwo, line 345: I wasn't really sure where this should be derived from.

Dirbaio:
> > Found waker not created by the Embassy executor. `embassy_time::Timer` only works with the Embassy executor.
> Could be related to the fact that I'm using a local version of embassy-time since using 0.4.0 won't build for me when also using a local version of embassy-stm32.

yes. make sure to use *all* embassy crates from local paths, do not mix crates.io and local.

charliefoxtwo:
> yes. make sure to use _all_ embassy crates from local paths, do not mix crates.io and local.

ok yeah that was the issue! Blinky example runs successfully now.

Dirbaio (approved):
thanks!

I've separated C0 RCC into two versions for the old/new as we discussed in Matrix.

charliefoxtwo:
Oh thanks! I've been so busy recently I haven't had time. Thank you for taking care of splitting that out. I'll pull it down and validate it still works.

PR #609, opened 2025-05-19
Add RTC support for h7rs

aaamorgan:
I tried to generally match the `rtc_v3.yaml` format where possible.

The embassy rtc example for stm32h7rs now works after these updates.


Dirbaio:
thanks!

PR #612, opened 2025-05-29
Add N6 support

tarfu:
Adding support for N6


Dirbaio:
looks good from a quick look. anything in particular you want me to check / any particular question?

tarfu:
If there are some unwritten naming conventions, you would probably spot them. The easiest is to look at the diff of https://github.com/embassy-rs/stm32-data/pull/612/commits/e2a5bdac64d9cb83d67332fbe514540e6c1ea165.

Fixing the duplication error I need to read up in the docs as in my understanding right now there is a bit of a difference in the fields flaged as duplicate.

tarfu:
As there is no flash I have not defined any flash if someone wonders at review.

Dirbaio:
bender run

tarfu:
Removed printing. They were nice to find some errors in the yaml.
Would think about adding log and with different levels, maybe sometime.

PR #613, opened 2025-06-03
feat(comp, g4): Add COMP pin definitions and register enums

IvanLi-CN:
Add pin definitions for COMP1-7 in data/extra/family/STM32G4.yaml. Add enum definitions and link them in data/registers/comp_v2.yaml.

### DIFF
--- a/data/extra/family/STM32G4.yaml
+++ b/data/extra/family/STM32G4.yaml
@@ -56,3 +56,73 @@ peripherals:
         signal: VP1
       - pin: PB13
         signal: VP2
+  - name: COMP1
+    pins:
+      - pin: PA1
+        signal: INP0
+      - pin: PB1
+        signal: INP1
+      - pin: PA4
+        signal: INM0
+      - pin: PA0
+        signal: INM1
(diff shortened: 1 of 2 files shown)


Dirbaio:
LOCK, SCALEN, BRGEN, EN enums should be deleted, they're redundant.

IvanLi-CN:
Sorry, I just noticed the notification email, please give me some time and I will correct the problem. At the moment I'm still trying to figure out what to do about embassy-stm32, when I make progress I'll change the code as you suggested and submit it.

leftger:
@IvanLi-CN It's been a year since your last comment but I still think it's a valid PR to bring in. If you can address dirbaio's comments and fix the merge conflicts I'm sure we can merge this promptly.

IvanLi-CN:
@leftger I’m really sorry, but I think I’ll have to abandon this PR. I originally planned to continue working on it after validating the implementation in my project, but I ended up not using COMP and completely forgot about it. With STM32G4 prices now several times higher, I’m unlikely to work with this series again anytime soon.

PR #614, opened 2025-06-03
Override pin names for OPAMPs

skoe:
Override pin names for OPAMPs

Some peripherals, such as OPAMPs, have PIN names and variants that are distributed across the chip XMLs and IP XMLs. It would be very cumbersome to reconstruct these completely by combining this information. If there are only a few versions of these IPs, it may make more sense to define them manually in YAML files.

This PR adds this mechanism and implements it consistently for all OPAMPs.

**Design decisions:**

* Instead of mixing the configuration from XML and manual additions, I have completely replaced the pins with YAML for transparency and clarity.

* In principle, the pins in all families could have been called either VPx/VMx or VINMx/VINPx. Instead, I used exactly the names from the RMs, i.e., VPx/VMx for F3, otherwise VINMx/VINPx.

* For devices without alternative OPAMP inputs, only VINM/VINP is used instead of VINM0/VINP0, which is also consistent with the RM.

A PR for embassy-stm32 that supports all of this is on the way.

**Implementation per family:**

| Family  | Example     | Old version     | Future Version | Perimap OK? | Pins OK? |
| ------- | ----------- | --------------- | ------- | ----------- | -------- |
| L1      | STM32L152RE | (tbd)           | v1      |     -       |    ?     |
| F3      | STM32F303K8 | opamp_f3.yaml   | v2      |     +       |    +     |
| L4      | STM32L432KC | opamp_l4u0.yaml | v3      |     +       |    +     |
| L4+     | STM32L4P5ZG | opamp_l4u0.yaml | v3      |     +       |    +     |
| H7      | STM32H755ZI | opamp_h.yaml    | v4      |     +       |    +     |
| L5      | STM32L552ZE | opamp_l4u0.yaml | v3      |     +       |    +     |
| G4      | STM32G474RE | opamp_g4.yaml   | v5      |     +       |    +     |
| U5      | STM32U545RE | opamp_l4u0.yaml | v3      |     +       |    +     |
| H5      | STM32H503RB | opamp_h.yaml    | v4      |     +       |    +     |
| U0      | STM32U031R8 | opamp_l4u0.yaml | v3      |     +       |    +     |
| U3      | STM32U385RG | opamp_l4u0.yaml | v3      |     +       |    +     |

'-' = not there, '?' = not checked, '+' = done

I'll rename \<old version\> to \<version\> soon, because in the driver code it is confusing to write things like `#[cfg(opamp_l4u0)]` if also other families are meant.


Dirbaio:
thanks! :rocket:

PR #623, opened 2025-06-17
Feat/support wba6xxx

leftger:
Added support for WBA65XX devices:

- Used RM0515 from ST to add new registers, enums.
- Used `chiptool` to parse out errors.

### DIFF
--- a/stm32-data-gen/src/interrupts.rs
+++ b/stm32-data-gen/src/interrupts.rs
@@ -502,7 +502,7 @@ fn valid_signals(peri: &str) -> Vec<String> {
         ("GTZC", &["GLOBAL", "ILA"]),
         ("WWDG", &["GLOBAL", "RST"]),
         ("USB_OTG_FS", &["GLOBAL", "EP1_OUT", "EP1_IN", "WKUP"]),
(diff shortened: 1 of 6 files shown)


stm32-data-gen/src/interrupts.rs
```diff
@@ -502,7 +502,7 @@ fn valid_signals(peri: &str) -> Vec<String> {
         ("GTZC", &["GLOBAL", "ILA"]),
         ("WWDG", &["GLOBAL", "RST"]),
         ("USB_OTG_FS", &["GLOBAL", "EP1_OUT", "EP1_IN", "WKUP"]),
-        ("USB_OTG_HS", &["GLOBAL", "EP1_OUT", "EP1_IN", "WKUP"]),
+        ("USB_OTG_HS", &["GLOBAL", "EP1_OUT", "EP1_IN", "WKUP", "USB"]),
```
Dirbaio, line 505: this is not quite right. If the chip has a single interrupt for USB, it should get the signal name "GLOBAL". Reasons:
- "USB" as a signal name is redundant because it's already known it's an interrupt for the USB peripheral.
- consistency between all families.

PR #625, opened 2025-06-24
Fixed clock source for SDMMC1 on STM32L433RC

SGINick:
Fixing the issue where the SDMMC clock source was running at 4Mhz when it was expected to run at 400KHz.  Original issue here:

https://github.com/embassy-rs/embassy/issues/4314

I still cannot get the SDMMC1 peripheral working, but the clock is definitely fixed, tested with an oscilloscope and it read a 2.49µs period.


Dirbaio:
thanks!

PR #626, opened 2025-06-25
ADC Support for STM32WBA series

purepani:
Adds support for the WBA series. 
I haven't yet tested this yet, so this will be put into draft mode.

### DIFF
--- a/stm32-data-gen/src/perimap.rs
+++ b/stm32-data-gen/src/perimap.rs
@@ -105,7 +105,7 @@ pub static PERIMAP: RegexMap<(&str, &str, &str)> = RegexMap::new(&[
     (".*:ADC:aditf5_40v2_U5", ("adc", "u5", "ADC")),
     (".*:ADC:aditf5_40v1_U5", ("adc", "u5", "ADC")),
     (".*:ADC:aditf4_v4_U5", ("adc", "u5", "ADC4")),
-    (".*:ADC:aditf4_v4_WBA_CUBE", ("adc", "wba", "ADC4")),
+    (".*:ADC:aditf4_v4_WBA_Cube", ("adc", "wba", "ADC4")),
     (".*:ADC:aditf5_v3_0_H7RS", ("adc", "h7rs", "ADC")),
     ("STM32WL5.*:ADC:.*", ("adc", "g0", "ADC")),
     ("STM32WLE.*:ADC:.*", ("adc", "g0", "ADC")),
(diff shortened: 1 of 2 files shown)


stm32-data-gen/src/perimap.rs
```diff
@@ -105,6 +105,7 @@ pub static PERIMAP: RegexMap<(&str, &str, &str)> = RegexMap::new(&[
     (".*:ADC:aditf5_40v2_U5", ("adc", "u5", "ADC")),
     (".*:ADC:aditf5_40v1_U5", ("adc", "u5", "ADC")),
     (".*:ADC:aditf4_v4_U5", ("adc", "u5", "ADC4")),
+    (".*:ADC:aditf4_v4_WBA_CUBE", ("adc", "wba", "ADC4")),
```
leftger, line 108: Sorry last thing.

The IP version shows up as `aditf4_v4_WBA_Cube`

I don't know if the regex distinguishes between capitalizations 🤷‍♂️

leftger (changes requested):
The one tiny change on `adc_wba.yaml` is really all I'd recommend to change. Thanks for doing this! 👍

purepani:
I fixed both the issues. However, I can't seem to get the peripheral to generate when running `./d gen-all`. Do you have any ideas why this may be?

leftger:
I'm looking at the problem rn myself too :)

I think it's the regex used to link the peripherals together from perimap.rs
I think the line was supposed to be this:     `(".*:ADC4:aditf4_v4_WBA_Cube", ("adc", "wba", "ADC4")),` on `perimap.rs`
You had `ADC:aditf4` I think it's supposed to be `ADC4:aditf4`

purepani:
There is one error when I run the `chiptool check` command, with an overlapping register, but it seemed like that register was able to be interpreted in 2 different ways depending on a different flag. I'm not sure whether that stopped it from generating or not.

stm32-data-gen/src/perimap.rs
```diff
@@ -105,6 +105,7 @@ pub static PERIMAP: RegexMap<(&str, &str, &str)> = RegexMap::new(&[
     (".*:ADC:aditf5_40v2_U5", ("adc", "u5", "ADC")),
     (".*:ADC:aditf5_40v1_U5", ("adc", "u5", "ADC")),
     (".*:ADC:aditf4_v4_U5", ("adc", "u5", "ADC4")),
+    (".*:ADC:aditf4_v4_WBA_Cube", ("adc", "wba", "ADC4")),
```
leftger, line 108: Replace this with:
```
    (".*:ADC4:ADC:aditf4_v4_WBA", ("adc", "wba", "ADC4")),
```

leftger (changes requested):
Just change that one line and it should all be good in the 'hood!

leftger:
Don't worry about the overlapping register. `adc_u5` has the same exact overlap so I don't see it as something inconsistent.

purepani:
Made the change; just making sure it generates on my end and ill push!

purepani:
Looks like it added the peripheral map, which wasn't there before, so I think it's good now.

leftger (approved):
Thanks so much!!!

PR #627, opened 2025-06-25
Fix WBA6x memory maps

bugadani:
RegexMap is unfortunately sensitive to ordering, so the new WBA6 data was ignored. There are also some suspicious L1 entries but I'll not even try to untangle that.


leftger:
Thank you for checking into this. I really appreciate it. 😄

PR #629, opened 2025-07-05
fix STM32F0 CRC peripheral version

martin2250:
the F07x and F09x series contain the more advanced version of the CRC peripheral

image

fixes https://github.com/embassy-rs/embassy/issues/4369


martin2250:
looking at the diff this is very much wrong...

martin2250:
the new diff looks fine

Dirbaio:
thanks!

PR #644, opened 2025-07-20
Added RNG support for STM32WBA devices

leftger:
- STM32WBA6 devices have 4 HTCR registers
- NSCR enum was missing
- Added new regex to handle STM32WBA devices

### DIFF
--- a/data/registers/rng_v3.yaml
+++ b/data/registers/rng_v3.yaml
@@ -13,10 +13,17 @@ block/RNG:
     description: data register
     byte_offset: 8
     access: Read
+  - name: NSCR
+    description: RNG noise source control register.
+    byte_offset: 12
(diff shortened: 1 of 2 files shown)


Dirbaio:
have you checked wither other rng_v3 chips have the regs you added? if not we need a new RNG version.

Dirbaio:
yeah if it's right in some devices and wrong in some others i'd just take it out and let the HAL deal with it...

leftger:
OK I added a new RNG for the WBA6 series of devices. I kept the enum for v3 since it's already being used.

Let me know if there's anything else I need to change @Dirbaio.

leftger:
Hey @Dirbaio is there anything else I need to do to this PR? I'm pretty sure I did what you asked.

PR #645, opened 2025-07-20
Fix WB55 SAI

tomaz-suller:
- **Copy `sai_v3_4pdm.yaml`**
- **Apply necessary changes**
- **Remove `NODIV` enum**
- **Update WB[35] family SAI mapping**

### DIFF
--- a/stm32-data-gen/src/perimap.rs
+++ b/stm32-data-gen/src/perimap.rs
@@ -204,6 +204,7 @@ pub static PERIMAP: RegexMap<(&str, &str, &str)> = RegexMap::new(&[
     (".*:SAI:sai1_v1_0", ("sai", "v1", "SAI")),
     (".*:SAI:sai1_v1_1", ("sai", "v2", "SAI")),
     (".*:SAI:sai1_v1_2", ("sai", "v2", "SAI")),
+    (r"STM32WB[35]5.*:SAI\d?:.*", ("sai", "v1_4pdm", "SAI")),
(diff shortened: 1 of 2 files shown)


stm32-data-gen/src/perimap.rs
```diff
@@ -204,6 +204,7 @@ pub static PERIMAP: RegexMap<(&str, &str, &str)> = RegexMap::new(&[
     (".*:SAI:sai1_v1_0", ("sai", "v1", "SAI")),
     (".*:SAI:sai1_v1_1", ("sai", "v2", "SAI")),
     (".*:SAI:sai1_v1_2", ("sai", "v2", "SAI")),
+    (r"STM32WB[35]5.*:SAI\d?:.*", ("sai", "v1_4pdm", "SAI")),
     (".*:SAI:sai1_v2_0", ("sai", "v1", "SAI")),
```
tomaz-suller, line 208: Since the `RegexMap` exits at the first match, we need to match the WB55 specifically before matching `sai1_v2_0`, since that is the version assigned incorrectly to the family by CubeMX.

@Dirbaio would it be okay to depend on this implementation detail of `RegexMap`? The alternative I see is changing the DB file.

tomaz-suller:
Another thing is perhaps we should drop the `NODIV` enum in a separate PR? This affects all SAI interfaces

stm32-data-gen/src/perimap.rs
tomaz-suller: Thanks, I'd missed that

PR #646, opened 2025-07-20
Fixed missing fields for TrustZone on WBA series

leftger:
- Part of a future addition for TrustZone support on WBA devices on Embassy
- Original work based off of WB devices, but more similar to U5


leftger:
I will definitely fix that. Here's a question:
Screenshot 2025-07-20 at 10 57 03 AM

Those missing enums were taken directly from the `flash_u5.yaml` file. Do you want me to fix the enum on that file too as part of the PR?

Dirbaio:
that'd be great yes :pray: 

ideally we should have no `B_0x*` enums anywhere, maybe we should add a check in CI for this

leftger:
OK the enums with missing names have been removed 😄

PR #648, opened 2025-07-20
rcc/wb: Expose the RFWKP clock mux for IPCC

clubby789:
This exposes `rfwkpsel` in `ClockMux` within `embassy-stm32`

Follow up to #643


Dirbaio:
reran the job, worked now

PR #649, opened 2025-07-21
Add STM32F1 AFIO remap

### LINKED ISSUE #10: Add stm32f1 GPIO AF maps

fwolter:
This adds the STM32F1 remap information for the AFIO_MAPR/AFIO_MAPR2 registers. The XML files only contain references to constants in the ST HAL. So, a custom lookup table was added.

Fixes #10


fwolter:
You're right, that structure makes more sence. I changed it.

PR #661, opened 2025-08-16
Add ASCR register to gpio_v2 for stm32l47x/stm32l48x

korbin:
Per the STM32L476 TRM:

> On STM32L47x/L48x devices, before any conversion of an input channel coming from GPIO pads, it is necessary to configure the corresponding GPIOx_ASCR register in the GPIO, in addition to the I/O configuration in analog mode.

### DIFF
--- a/data/registers/gpio_v2.yaml
+++ b/data/registers/gpio_v2.yaml
@@ -42,6 +42,10 @@ block/GPIO:
       stride: 4
     byte_offset: 32
     fieldset: AFR
+  - name: ASCR
+    description: GPIO port analog switch control register
+    byte_offset: 44
+    fieldset: ASCR
 fieldset/AFR:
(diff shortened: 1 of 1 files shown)


korbin:
> we can't simply add this to gpio_v2.yaml, that'd make it available on all chips using that version. It should be a new gpio version.

what should it be called? AFAIK, it's just these two chips in the series and the rest of the registers are the same.

Dirbaio:
`gpio_v2_l478` probably.

Dirbaio:
with [this commit](https://github.com/embassy-rs/embassy/pull/4620/commits/e9c25ea8b7fdf921e5988eceab6354e557b028d5) underscores in versions are treated as "sub-versions", so `gpio_v2_l478` will enable both cfgs `gpio_v2_l478` and `gpio_v2` so it should work out of the box.

PR #662, opened 2025-08-20
stm32l4: enable comparator support

hvraven:
The bits for `COMP1` exactly match those in the existing definition of `comp_v3`. `COMP2` has an extra bit `WINMODE`, used to connect the positive inputs of both comparators. This is not described at the moment. I was not sure how this should be described, and don't need it at the moment.

The used register definition is lacking the description for the `INxSEL` and `INMESEL` registers. These also have variations between the different comparator units, as they select the GPIO pins used. There also appears to be an error in the reference manual ([RM0394](https://www.st.com/resource/en/reference_manual/rm0394-stm32l41xxx42xxx43xxx44xxx45xxx46xxx-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) on page 538,  as for `COMP1` the description of `INMSEL` states that `0b110` uses `INMESEL` for selection, whereas the description in the `INMESEL` register states that `0b111` should be used. The later would be in line with `COMP2`. I currently don't have an easy way to check which one is correct. It also doesn't affect this change, as the enums are not described.

### DIFF
--- a/stm32-data-gen/src/perimap.rs
+++ b/stm32-data-gen/src/perimap.rs
@@ -599,6 +599,7 @@ pub static PERIMAP: RegexMap<(&str, &str, &str)> = RegexMap::new(&[
     ("STM32G4.*:.*:COMP:.*", ("comp", "v2", "COMP")),
     ("STM32U0.*:.*:COMP:.*", ("comp", "u0", "COMP")),
     ("STM32WL.*:.*:COMP:.*", ("comp", "v3", "COMP")),
+    ("STM32L4.*:.*:COMP:.*", ("comp", "v3", "COMP")),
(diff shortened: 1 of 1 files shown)


hvraven:
There is an actual difference, probably requiring an independent copy of the register definitions. `PWRMODE` does not support low speed in this variant, the bits also map to medium speed. 
Looking at the other definitions, which naming scheme is preferred for register definitions? Some use `Pwrmode` (following the ST register naming), while others use full names, such as `PowerMode`.

With some guidance on the differences between `COMP1` and `COMP2` (only the extra bit in `COMP2`), I will prepare a new PR with an independent definition if desired. Until then I would keep it open, as it is not really wrong, and an easy fix to make the registers accessible in the STM32L4 family

PR #663, opened 2025-08-20
Update I2S registers in spi_v2 to be consistent with spi_v3

pthien:
[Problem]
spi_v2 hardware supports I2S, but it has not been enabled in the embassy HAL. Some of the fields in the I2SCFGR register for spi_v2 seem to be incorrect compared to spi_v3

[Solution]
- Make the enum for the `I2SCFG`, `I2SSTD` fields match the field name - they were missing the '2'
- Remove the enum for `I2SMOD`. It is not present in spi_v3 and falls under the category of "useless enums"

### DIFF
--- a/data/registers/spi_v2.yaml
+++ b/data/registers/spi_v2.yaml
@@ -197,7 +197,7 @@ fieldset/I2SCFGR:
     description: I2S standard selection
     bit_offset: 4
     bit_size: 2
-    enum: ISSTD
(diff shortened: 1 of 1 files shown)


i509VCB (approved):
I don't usually merge stm32 stuff, but I have verified the similarities and checked if G0B1 and C071 (spi_v2) match.

PR #665, opened 2025-09-02
stm32u0: Add UEDE field to LPTIMx_DIER register

3j14:
Added the UEDE ("Update event DMA request enable") single bit field to lptim_v2b.yaml, valid for all STM32U0 parts.

See reference manual RM0503 (rev. 4) section 27.7.5.


3j14:
Thanks! 😄

PR #671, opened 2025-09-15
Add BKPSRAM to H5

usbalbin:
Add battery backed sram regions for stm32h5 family


stm32-data-gen/src/memory.rs
```diff
@@ -228,11 +228,11 @@ static MEMS: RegexMap<&[&[Mem]]> = RegexMap::new(&[
                                         mem!(BANK_1 { 0x08000000 256 }, BANK_2 { 0x08040000 256 }, CCMRAM_ICODE { 0x10000000 32 }, SRAM1 { 0x20000000 80 }, SRAM2 { 0x20014000 16 }, CCMRAM_DCODE { 0x20018000 32 })
                                       ]),
```
usbalbin, line 229: I had a quick loog at the G4 family and that seems to have 32 32bit registers (a total of 128 bytes) as part of the `TAMP` register block. Does that affect anything of what we want to do here?

PR #675, opened 2025-09-24
Add support for I2Sext peripherals on stm32f3xx / stm32f4xx

bifurcation:
Full-duplex I2S opeeration on these platforms requires "I2Sext" peripherals that work alongside the SPI/I2S peripherals.  This PR attempts to add support for these peripherals, in support of [enabling full-duplex operation at the `embassy-stm32` layer](https://github.com/embassy-rs/embassy/issues/4698).

I tried following the directions in the README, invoking `./d extract-all` for `I2S2ext` and `I2S3ext`.  This produced a few dozen YAML files, of which there were two unique classes modulo renaming of fields.  I called these `i2sext_v1.yaml` for the F4 family and `_v2.yaml` for the F3 family (following release date order).  I added lines to `perimap.rs` which ought to have caught the relevant peripherals.

However, I hit a roadblock here.  The entries in `PERIMAP` never hit, apparently because the I2Sext peripherals don't appear in the chip XML in `data/cubedb`, from which the `ChipGroup` values are generated.  Those XML files *do* contain the pin definitions that the ext-peripherals require (e.g., in `STM32F405RGTx.xml`, pin `PB4` has `I2S3_ext_SD`).

I would appreciate guidance on how to move forward here.  I suspect that the I2Sext peripherals might need special handling like the I2S peripherals, but this is exceeding the bounds of my knowledge here.

### DIFF
--- a/data/registers/i2sext_v1.yaml
+++ b/data/registers/i2sext_v1.yaml
@@ -1,0 +1,681 @@
+block/SPI1:
+  description: Serial peripheral interface.
+  items:
+  - name: CR1
+    description: control register 1.
+    byte_offset: 0
+    fieldset: CR1
+  - name: CR2
+    description: control register 2.
+    byte_offset: 4
+    fieldset: CR2
+  - name: SR
+    description: status register.
+    byte_offset: 8
+    fieldset: SR
+  - name: DR
+    description: data register.
+    byte_offset: 12
+    fieldset: DR
+  - name: CRCPR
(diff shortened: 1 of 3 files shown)


Dirbaio:
I don't think I2SExt is a "real" peripheral though? just a mode of i2s?

xoviat:
There may be extra registers here, not sure.

xoviat:
Only the extra registers should be added in the i2s yaml.

PR #685, opened 2025-11-20
initial stm32u3 support

graynode:


data/registers/rcc_u3.yaml
```diff
@@ -0,0 +1,2600 @@
...
+          value: 3
+enum/I2CSEL:
+    bit_size: 1
+    variants:
+        - description: PCLK1 selected.
+          name: PCLK1
+          value: 0
+        - description: MSIK selected.
+          name: MSIK
+          value: 1
```
xoviat: Any enums that have the same variants need to be combined.

data/registers/rcc_u3.yaml
```diff
@@ -0,0 +1,2600 @@
...
+          name: GPIOCSLPEN
+        - bit_offset: 3
+          bit_size: 1
+          description: I/O port i clock enable during Sleep mode (i = E to A).
+          name: GPIODSLPEN
+        - bit_offset: 4
+          bit_size: 1
+          description: I/O port i clock enable during Sleep mode (i = E to A).
+          name: GPIOESLPEN
+        - bit_offset: 6
```
xoviat: This needs to be combined into an array.

graynode: It's already an array similar to https://github.com/embassy-rs/stm32-data/blob/903d26b1c3dbe923bcda24ef91031876eaa402be/data/registers/rcc_u5.yaml#L405 the fields are just a list of dictionaries. Again sorry for not fully understanding.

data/registers/rcc_u3.yaml
```diff
@@ -0,0 +1,2600 @@
...
+          description: I/O port i clock enable during Stop mode (i = E to A).
+          name: GPIOESTPEN
+        - bit_offset: 6
+          bit_size: 1
+          description: I/O port i clock enable during Stop mode (i = H to G).
+          name: GPIOGSTPEN
+        - bit_offset: 7
+          bit_size: 1
+          description: I/O port i clock enable during Stop mode (i = H to G).
+          name: GPIOHSTPEN
```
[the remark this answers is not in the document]
graynode: Same as above response. I believe this is already is following the format of list of dictionaries. Similar to rcc_u5.yaml

xoviat:
You're right, I went back and looked at the other files, and they are like this.

stm32-data-gen/src/interrupts.rs
```diff
@@ -524,7 +524,10 @@ fn valid_signals(peri: &str, chip_name: &str) -> Vec<String> {
         ("I2C", &["ER", "EV"]),
         ("I3C", &["ER", "EV", "WKUP"]),
         ("FMPI2C", &["ER", "EV"]),
-        ("TIM", &["BRK", "UP", "TRG", "COM", "CC"]),
+        (
+            "TIM",
+            &["BRK_TERR_IERR", "BRK", "UP", "TRG", "TRG_COM_DIR_IDX", "COM", "CC"],
```
Dirbaio: `BRK_TERR_IERR`, `TRG_COM_DIR_IDX` don't look like signal names. They're most likely a combination of multiple signals, where the signals are actually `BRK`, `TERR`, `IERR`, etc.

No idea what these are (TERR, IERR, DIR, IDX). Maybe they're something new in U3? If so we should ideally make it so they only appear in the U3 generated JSONs and not other families. 

If you want to just those signals for now that's fine too, they can be added later when someone needs them.

PR #705, opened 2025-12-06
Add UID and DESIG blocks to WBA series

leftger:
Refactor: restored VREFINTCAL for WBA series

Fixed DESIG base address from imported headers

Fixed DESIG generation to only gen for WBA series


leftger:
OK I think I have it. Apologies if anyone got a crazy amount of notifications. I was trying to not break the previous VREFINTCAL generation but also allow the header files to distinguish between the WBA5 and the WBA6 families to generate the DESIG peripheral at the right address.

xoviat:
Can you merge upstream?

leftger:
Let me do that real quick

leftger:
OK I merged upstream :)

PR #706, opened 2025-12-06
Add STM32WL3xx variants

marcemmers:
The SVD data in https://github.com/embassy-rs/stm32-data-sources/pull/14 is quite poor, so there are not a lot of enum definitions I am afraid.


leftger:
@marcemmers have you found better SVD sources? You might be able to add them to `stm32-data-sources`, this repo's sister project.

PR #729, opened 2026-02-09
STM32N6: Add the UART peripherals

lsartory:
Add the UART, USART and LPUART peripherals for the STM32N6.
Improvement related to the clock settings and multiplexers are also included.


lsartory:
Thanks!

PR #741, opened 2026-02-26
Bring comp_v2 closer to comp_u5

usbalbin:
Since u5 seems to be the only comp with support in embassy-stm32, I tried to make comp_v2 to be as close as possible to that

### DIFF
--- a/data/registers/comp_v2.yaml
+++ b/data/registers/comp_v2.yaml
@@ -16,6 +16,7 @@ fieldset/CSR:
     description: Comparator signal selector for inverting input INM. (RM0440 24.3.2 Table 197)
     bit_offset: 4
     bit_size: 3
+    enum: INM
   - name: INPSEL
     description: Comparator signal selector for non-inverting input INP. (RM0440 24.3.2 Table 196)
     bit_offset: 8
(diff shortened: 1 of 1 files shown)


usbalbin:
Is there anything more that to needs to be done here before Comp support can be added to embassy-stm32?

I get errors about when trying to enable src/comp.rs for g4.

```
no associated item named `RCC_INFO` found for struct `peripherals::COMP1` in the current scope
```

I have not quite wrapped my head around how everything works together in embassy w.r.t. the pac and hal and all the magic build scripts :)

xoviat:
It looks like RCC info is missing for comp:

https://github.com/embassy-rs/stm32-data-generated/blob/3f3a1b8fc72fbbbcea64b88018ae5ab83ce7aeb0/data/chips/STM32G484CE.json#L2595

Here is what RCC info should look like:

https://github.com/embassy-rs/stm32-data-generated/blob/main/data/chips/STM32G484CE.json#L902

RCC info is required if the peripheral is clocked by an hclk or a pclk. From the refrence manual:

> The COMP clock provided by the clock controller is synchronous with the APB2 clock.
There is no COMP-dedicated clock enable control bit in the RCC controller. Reset and clock
enable bits are common for COMP and SYSCFG

For this description, 

```
            "bus_clock": "PCLK2",
            "kernel_clock": "PCLK2",
            "enable": {
              "register": "APB2ENR",
              "field": "SYSCFGEN"
            },
```

stm32-data-gen could be modified to copy the rccinfo from the syscfg for comp and remove the reset bit.

xoviat:
The reset bit is optional and must be removed so that syscfg is not reset:

```
            pub struct Rcc {
                pub bus_clock: String,
                pub kernel_clock: rcc::KernelClock,
                pub enable: rcc::Field,
                #[serde(skip_serializing_if = "Option::is_none")]
                pub reset: Option<rcc::Field>,
                #[serde(default, skip_serializing_if = "crate::is_default")]
                pub stop_mode: rcc::StopMode,
            }
```

PR #745, opened 2026-03-02
Add dac inc triggers

usbalbin:


usbalbin:
Oh, sorry. I have tested this with https://github.com/embassy-rs/embassy/pull/5411 and a HRTIM setup and it seems to work fine so far, producing a sawtooth waveform synced to the pwm output

(The pwm frequency is about 250kHz som the DAC is probably a bit too slow to get sharp edges)
PXL_20260303_105553373 MP

Issue #594, opened 2025-04-22
STM32WB0 series support


rick-de-water:
After prototyping with this crate on a blue pill (STM32F103C8), I started looking for the specific STM32 I actually want to use, and I ended up deciding to go with the STM32WB09KE. I got myself a [Nucleo dev board](https://www.st.com/en/evaluation-tools/nucleo-wb09ke.html) and started updating my code, only to find out that the entire [STM32WB0 series](https://www.st.com/en/microcontrollers-microprocessors/stm32wb0-series.html) is not supported yet.

I wouldn't mind contributing, but I'm also quite new to the embedded world. Would it be difficult to add support?

PR #746, opened 2026-03-02
STM32WB0 - Perimap Corrections & Additional Peripherals

physgun:
Fixes incorrect perimap.rs assignments and add missing peripherals for the STM32WB0 family: 05, 06, 07, & 09. Part of an effort to address #594.

References used:
- RM0529 Rev 4 - STM32WB05
- RM0530 Rev 3 - STM32WB06 & 07 (both chips are identical in every way, save for the total amount of RAM)
- RM0505 Rev 4 - STM32WB09

Changes:
- [x] GPIO switched to gpio_v2 to match registers.
- [x] ADC corrected to match to adc_wb0.yaml.
- [x] New TRNG peripheral, trng_v1.yaml, for WB09, replaces RNG
    - [x] Added transform TRNG.yaml to remove TRNG_ prefixes.
- [x] Added new RADIO peripheral, radio_v1.yaml. Not yet tested. 
- [x] PWR corrected to match pwr_wb0.yaml
    - [x] Needed to be extracted as "PWRC".
- [x] CRC version corrected to v3.
- [x] TIMx changed to.. what I think are the correct types?  Please advise if not.
- [x] Added FLASH peripheral.
- [x] Added SYSCFG peripheral. 
- [x] Added RRM peripheral.
- [x] Fixed bug in generator.rs where apply_family_extras would overwrite peripherals if a chip family YAML specified a peripheral with a non-zero address and also had no pins.
- [x] Added register merging to apply_family_extras, if the extra has Some registers specified and the generated has None.

### DIFF
--- a/data/extra/STM32WB0.yaml
+++ b/data/extra/STM32WB0.yaml
@@ -1,4 +1,5 @@
----
+matches:
+  family: STM32WB0
 peripherals:
   - name: SYSCFG
     address: 0x40000000
--- a/stm32-data-gen/src/generator.rs
+++ b/stm32-data-gen/src/generator.rs
@@ -792,19 +792,20 @@ fn apply_extras(
     peripherals: &mut HashMap<String, stm32_data_serde::chip::core::Peripheral>,
 ) {
     for extra in extras {
-        let family_matches = extra.matches.family.as_ref().map_or(true, |f| f == &group.family);
+        let family_matches = extra.matches.family.as_ref().is_none_or(|f| f == &group.family);
         let chip_matches = extra
             .matches
             .chip
             .as_ref()
-            .map_or(true, |c| Regex::new(&format!("^{c}$")).unwrap().is_match(chip_name));
+            .is_none_or(|c| Regex::new(&format!("^{c}$")).unwrap().is_match(chip_name));
         if !family_matches || !chip_matches {
             continue;
         }
 
         // merge extra peripherals
-        if let Some(extras) = extra.peripherals {
-            for mut extra_periph in extras {
+        if let Some(extras) = &extra.peripherals {
+            for extra_periph in extras {
+                let mut extra_periph = extra_periph.clone();
                 let Some(generated_periph) = peripherals.get_mut(&extra_periph.name) else {
                     if extra_periph.address != 0 {
                         // filter out pins that may not exist in this package.


physgun:
I'd like some some guidance from maintainers on a few roadblocks I've hit:

- **WB05 & 09 / WB06 & 07 split.** The WB0 chip family has some diverging functionality between 05/09 and 06/07, leading to split .yamls between the two, if the PAC should never access reserved fields. I can merge the two if it is the HAL's responsibility to segregate the two groups and not the PAC's.
- **Useless Enums.** What is the criteria for a useless enum? Is it just any enum that generates only two variants, even if the descriptions might be useful? Can a bigger enum be useless? Should I just delete all of the enums and have the HAL deal with it? Some of the newer peripherals came with no enums at all.
- **pac.rs ```pub const PERIPHERAL1```.** The generated pac.rs for all chips has LPUART1 and USART1 ```pub const``` names for those peripherals. Is there somewhere to change those to be generated without the 1, or does it need to be manually edited out?
- **perimap.rs cannot find SYSCFG.** WB0's SYSCFG peripheral is not generating, and I haven't found out why. The peripheral is called SYSTEM_CTRL in the SVD, and there are two lines in the .xml that reference syscfg, "SYS" and "DEBUG", but trying to regex match those isn't working. What am I missing?
- **FLASH peripheral not in MCU .xml.** It's called FLASH_CTRL in the SVD, but I can't find any references to it in the .xml files for all chips. What do?

xoviat:
>  I can merge the two if it is the HAL's responsibility to segregate the two groups and not the PAC's.

Let's merge them so that the PAC size is not increased. 

> Useless Enums. What is the criteria for a useless enum? Is it just any enum that generates only two variants, even if the descriptions might be useful? Can a bigger enum be useless? Should I just delete all of the enums and have the HAL deal with it? Some of the newer peripherals came with no enums at all.

1. Convert ENABLE/DISABLE to bool. 
2. Merge equivalent enums.

> pac.rs pub const PERIPHERAL1. The generated pac.rs for all chips has LPUART1 and USART1 pub const names for those peripherals. Is there somewhere to change those to be generated without the 1, or does it need to be manually edited out?

This should not be changed to maintain consistent naming across all chips.

> FLASH peripheral not in MCU .xml. It's called FLASH_CTRL in the SVD, but I can't find any references to it in the .xml files for all chips. What do?

Add to `extra/family`

stm32-data-gen/src/perimap.rs
```diff
@@ -577,11 +581,11 @@ pub static PERIMAP: RegexMap<(&str, &str, &str)> = RegexMap::new(&[
     ("STM32U0.*:DBGMCU:.*", ("dbgmcu", "u0", "DBGMCU")),
     ("STM32WB0.*:DBGMCU:.*", ("dbgmcu", "wb0", "DBGMCU")),
     ("STM32U3.*:DBGMCU:.*", ("dbgmcu", "u3", "DBGMCU")),
+    ("STM32WB0.*:DBGMCU:.*", ("dbgmcu", "wb0", "DBGMCU")),
     ("STM32WBA.*:DBGMCU:.*", ("dbgmcu", "wba", "DBGMCU")),
-    ("STM32WB.*:DBGMCU:.*", ("dbgmcu", "wb", "DBGMCU")),
+    ("STM32WB[^0].*:DBGMCU:.*", ("dbgmcu", "wb", "DBGMCU")),
     ("STM32WL.*:DBGMCU:.*", ("dbgmcu", "wl", "DBGMCU")),
     ("STM32F1.*:GPIO.*", ("gpio", "v1", "GPIO")),
-    ("STM32WB0.*:GPIO.*", ("gpio", "v1", "GPIO")),
```
xoviat: Was this intentional?

physgun: The DBGMCU change was not intentional, the GPIO removal was intentional.

xoviat:
bender run

PR #775, opened 2026-04-15
gen: decompose GTZC into typed sub-peripherals (TZSC, TZIC, MPCBB)

leftger:
Decomposes the GTZC (Global TrustZone Controller) monolithic peripheral into typed sub-peripherals — GTZC_TZSC (security configuration), GTZC_TZIC (interrupt controller), and GTZC_MPCBB1/2/3/6 (block-based memory protection) — for the L5, H5, U5, and WBA chip families.

## Register definitions

- **New `gtzc_l5.yaml`**: L5-specific TZIC block (3-word IER/SR/FCR arrays) and MPCBB block (VCTR/LCKVTR), which differ structurally from the v1/wba variants.
- **Extended `gtzc_v1.yaml` / `gtzc_wba.yaml`**: Added `block/TZIC` and `block/MPCBB` definitions (previously only GTZC_TZSC existed).

## Peripheral mapping

- Adds `perimap.rs` entries for `GTZC_TZIC` and `GTZC_MPCBB*` sub-peripherals on H5, U5, WBA, and L5.
- Fixes L5 mapping: was incorrectly pointing at the WBA variant; now has its own `l5` version.

## Generator

- Adds `GTZC`, `GTZC_TZSC`, `GTZC_TZIC`, `GTZC_MPCBB1/2/3/6` to the ghost peripherals list so their base addresses are resolved from C headers.
- Soft-fail (warn and skip) when a chip has no C header, instead of panicking.

## Interrupt routing

TrustZone chips (L5, H5, U5, WBA) use NVIC2 (non-secure) as their primary NVIC, but `GTZC_IRQn` is defined only in NVIC1 (secure). The fix merges both NVICs for single-core TrustZone chips while explicitly excluding multi-core chips (STM32H7xx, STM32WL5xx) where NVIC1 and NVIC2 belong to different CPU cores and must not be mixed.

Additionally, the NVIC XML uses `GTZC_S` / `GTZC_NS` as the IP instance name in interrupt strings, which did not match the generated peripheral name `GTZC_TZIC`. A remapping is added so `GTZC_IRQn` is correctly attributed to `GTZC_TZIC`.

Bug fixes made necessary by the NVIC merge:
- **RTC signals**: Restrict the "prefer peri-name IRQ" dedup to `GLOBAL` signals only, preventing `RTC_Alarm` and `TAMPER` from being incorrectly collapsed to `RTC`.
- **LTDC UP signal**: Scope to STM32N6 only via `CHIP_SPECIFIC_SIGNALS`; other chips only have `ER`/`LO`.
- **HSEM1/HSEM2**: Map both names to `("HSEM", "GLOBAL")` for dual-core chips where each core has its own HSEM IRQ number.
- **F100 DMA filter**: Use `chip_name.starts_with("STM32F100")` instead of NVIC version string comparison.

### DIFF
--- a/data/registers/gtzc_l5.yaml
+++ b/data/registers/gtzc_l5.yaml
@@ -13,10 +13,6 @@ block/GTZC_TZSC:
     description: GTZC1 TZSC secure configuration register 2.
     byte_offset: 20
     fieldset: TZSC_SECCFGR2
-  - name: TZSC_SECCFGR3
-    description: GTZC1 TZSC secure configuration register 3.
-    byte_offset: 24
-    fieldset: TZSC_SECCFGR3
   - name: TZSC_PRIVCFGR1
     description: GTZC1 TZSC privilege configuration register 1.
     byte_offset: 32
@@ -25,10 +21,6 @@ block/GTZC_TZSC:
     description: GTZC1 TZSC privilege configuration register 2.
     byte_offset: 36
     fieldset: TZSC_PRIVCFGR2
-  - name: TZSC_PRIVCFGR3
-    description: GTZC1 TZSC privilege configuration register 3.
-    byte_offset: 40
-    fieldset: TZSC_PRIVCFGR3
 fieldset/TZSC_CR:
   description: GTZC1 TZSC control register.
   fields:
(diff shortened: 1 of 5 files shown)


leftger:
I neglected to link up some interrupt signals. I need to work on this some more.

leftger:
OK it's ready now.

PR #778, opened 2026-04-23
fix: stm32u3 memory

donRaphaco:
Fixes for stm32 flash in memory.rs

- flash BANK_2 base address 0x08080000 for 1MB and 0x08040000 for 512KB flash variants
- erase size for stm32u3 is 4096K


donRaphaco:
I'm planing on implementing the flash driver for the u3 on embassy-stm32 and hope this gets accepted first :)

PR #779, opened 2026-04-24
STM32N6: add sdmmc and camera/csi-2 peripherals

tarfu:
# STM32N6: SDMMC, MIPI CSI-2, DCMIPP, and RCC fixes

Adds the missing peripherals for SD-card, MIPI camera, and camera-pipeline support on STM32N6, plus two related RCC fixes.

## SDMMC v3 (`sdmmc2_v3_0_Cube`)

N6's SDMMC isn't register-compatible with the existing `sdmmc/v2`: the IDMA block is redesigned.
Core state-machine registers (POWER, CLKCR, CMDR, DCTRL, STAR, ICR, MASKR, ACKTIMER, IDMACTRLR, FIFOR) are unchanged.

Adds `data/registers/sdmmc_v3.yaml`, `transforms/SDMMC.yaml`, and the perimap rule.

## MIPI CSI-2 and DCMIPP

- `data/registers/csi_v1.yaml` (46 regs) + `transforms/CSI.yaml`, perimap `CSI:v1_0 → csi/v1`.
- `data/registers/dcmipp_v2.yaml` (240 regs) + `transforms/DCMIPP.yaml`, perimap `DCMIPP:cci_v2_0 → dcmipp/v2`. Named `v2` to match ST's cube IP version; `dcmipp_v1` is left reserved for the distinct H7RS variant (not in this PR).
- PSSI perimap regex extended to include N6 (`pssi_v1.yaml` already fits; cube version `cci_v3_0_Cube` same as H5).

## RCC fixes

**Revert DCMIPP→DCMI rename in `rcc_n6.yaml`.**  This restores the original names and adds the missing `CCIPR1.DCMIPPSEL` (kernel-clock mux).

**Extend the `DCMI`/`PSSI` fallback table** in `stm32-data-gen/src/rcc.rs`:

Comments in Code were generated with Claude to have them more descriptive.

PR #782, opened 2026-05-03
STM32N6: Add VENC and DLYB_SDMMC1/2 peripherals

tarfu:
Follow-up to #779.

- **VENC** (H.264/JPEG encoder)
- **DLYB_SDMMC1/2**: Looks different when looking at the svd but headers match existing peripheral. Added old dlyb to N6 from H7.

Disclosure: as always, code comments were writte/enhanced by claude.

### DIFF
--- a/data/extra/STM32N6.yaml
+++ b/data/extra/STM32N6.yaml
@@ -103,3 +103,21 @@ peripherals:
       kind: risaf
       version: n6
       block: RISAF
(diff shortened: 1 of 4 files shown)


tarfu:
@xoviat Sorry for the PR spam :D I'm porting stuff to the HAL from a prototype project the last few month.

PR #787, opened 2026-05-11
Add stm32c5 rcc

usbalbin:
Also see https://github.com/embassy-rs/stm32-data-sources/pull/18


xoviat:
Names need to be fixed so that embassy-stm32 can parse them correctly.

data/registers/rcc_c5.yaml
```diff
@@ -1,2646 +1,2183 @@
...
     value: 0
-  - name: B_0x1
-    description: PSI nominal frequency = 144 MHz.
+  - name: F144MHz
+    description: PSI nominal frequency = 144 MHz
     value: 1
+  - name: F160MHz
+    description: PSI nominal frequency = 144 MHz
+    value: 2
+  - name: F160MHz_other
```
usbalbin: Any suggestions for naming this?

data/registers/rcc_c5.yaml
```diff
@@ -1,2646 +1,2183 @@
...
   bit_size: 4
   variants:
-  - name: B_0x0
-    description: HSI clock divided by 1.
+  - name: Div1
+    description: HSI clock divided by 1
     value: 0
-  - name: B_0x1
-    description: HSI clock divided by 1.
+  - name: Div1_
```
usbalbin: Any suggestions for naming this?

xoviat:
@usbalbin do you have a c5 that you are looking to get working? If so, then I can help with this.

usbalbin:
I have ordered a [nucleo-c5a3](https://www.st.com/en/evaluation-tools/nucleo-c5a3zg.html), should pop up in the next week or so. But I have no concrete plans for what to do with it other than the feeling that the C5 seems really competetive for the price. So having support in embassy could probably come in handy at some point. :)

Some help or guidance would be much appreciated :)

---

Some general questions

1. Is there some goto way of naming mutiple enum variants that do the same thing? For example, we have enum `HSIKDIV` with `Div1` and `Div1_`. We have enum `PSIFREQ` with `F160MHz` and `F160MHz_other` etc.
2. CI errors on
    > Error: rcc: prohibited variant name PSIK in enum USARTSEL for rcc_c5
  
    Should I add "PSIK" to `allowed_variants` of stm32-data-gen/src/rcc.rs:56? Or should it be `PSI` which already exists

xoviat:
> Is there some goto way of naming mutiple enum variants that do the same thing? For example, we have enum HSIKDIV with Div1 and Div1_. We have enum PSIFREQ with F160MHz and F160MHz_other etc.

Is there a reason for the other variant? If there's no reason for the other variant, then can we just remove it?

> Should I add "PSIK" to allowed_variants of stm32-data-gen/src/rcc.rs:56? Or should it be PSI which already exists 

These are supposed to be the clock values accessible in `get_freqs()` in embassy-stm32. You would need to look at the clock tree in CubeMX to confirm this.

usbalbin:
Clock tree for the C5A3 (apparently there is now a Cube MX **2**)

C5A3-clock-tree

usbalbin:
I was too slow :P

That should be it now

usbalbin:
Also note the SVD PR here https://github.com/embassy-rs/stm32-data-sources/pull/18

PR #802, opened 2026-05-25
Add RNG peripheral for N6

william-spinelli:


william-spinelli:
@xoviat Sorry for tagging you directly.  But you seem to be very active also on N6 lately.
I also have a test example for rng that i can add to embassy once RNG peripheral gets added to metapac

PR #854, opened 2026-06-12
Fix incorrect OCTOSPI2, SDMMC, DSI requests for h7 chips

### LINKED ISSUE #6309: OCTOSPI2 async/MDMA doesn't seem to work on stm32h723vg
When I run
```rust
let mut jedec = [0u8; 3];
flash.ospi.read(
    &mut jedec,
    TransferConfig {
        iwidth:      OspiWidth::SING,
        instruction: Some(0x9F),
        isize:       AddressSize::_8Bit,
        dwidth:      OspiWidth::SING,
        ..Default::default()
    },
).await.unwrap();
```
the code hangs on the read and doesn't return. Likewise `blocking_read_dma` does not return either. However, `blocking_read` works fine, so my guess is something to do with MDMA. Using `OCTOSPI1` instead of `OCTOSPI2` does actually return, but then I run into #4069 and it returns all zeros.

I set up the bus in a configuration [added in this PR](https://github.com/embassy-rs/embassy/pull/6100):
```rust
let ospi_raw = Ospi::new_quadspi(
    p.OCTOSPI2,
    p.PB2,  // CLK
    p.PE7,  // D0
    p.PE8,  // D1
    p.PE9,  // D2
    p.PE10, // D3
    p.PE11, // NCS
    p.MDMA_CH1,
    Irqs,
    ospi_config(),
);
```
Irqs:
```rust
bind_interrupts!(struct Irqs {
    MDMA => dma::InterruptHandler<MDMA_CH1>;
});
```

I am not sure if this is set up correctly; there is no ospi mdma async example that I could find.

Summary:
- Blocking OCTOSPI1: all zeros #4069
- Async MDMA OCTOSPI1: all zeros #4069
- Blocking OCTOSPI2: works fine
- Async MDMA OCTOSPI2: hangs forever -- `blocking_read_dma` does not work either.

serena-proception:
Cross-reference STM's C headers: https://github.com/STMicroelectronics/stm32h7xx-hal-driver/blob/7a9433030448812bf845dd38ea6c8e71e609c565/Inc/stm32h7xx_hal_mdma.h#L327

DSI module is not present on these chips so trigger IDs 27 and 28 should be skipped. I also updated some SDMMC triggers because they appeared to be incorrect.

Claude Cowork found the incorrect configuration. I manually tested on an stm32h723vg with a W25Q512JVEIQ flash chip and this fix resolves https://github.com/embassy-rs/embassy/issues/6309. I was unable to test SDMMC or DSI because I don't have a board wired for that, but I changed it to be the same as the ST headers.


xoviat:
Thanks! I'm not sure whether there's a way to verify this data in general, but we can add tooling to verify the request numbers from these repositories, if you think it's worth it. Of course claude is another option, but any improvements that you think would improve data integrity are welcome.

serena-proception:
> Thanks! I'm not sure whether there's a way to verify this data in general, but we can add tooling to verify the request numbers from these repositories, if you think it's worth it. Of course claude is another option, but any improvements that you think would improve data integrity are welcome.

My idea would be to run the C preprocessor and cross-validate the yaml files with the #DEFINEs. I dont really have time to work on this right now though

Claude found this header file and noticed the discrepancy, but it didn't patch it correctly. But it saved me a lot of time in debugging and narrowing down the problem.

PR #869, opened 2026-06-27
C5 flash cleanup

usbalbin:


data/registers/flash_c5.yaml
```diff
@@ -153,28 +153,27 @@ block/FLASH:
...
     access: Read
-    fieldset: WRP2R_CUR
+    fieldset: WRP2R
   - name: WRP2R_PRG
     description: FLASH write page protection for bank2.
     byte_offset: 492
-    fieldset: WRP2R_PRG
-  - name: HDP2R_CUR
+    fieldset: WRP2R
+  - name: HDP2R
```
usbalbin: oh, that should not be changed

PR #882, opened 2026-07-26
C5 uses H5's AES/SAES

usbalbin:
Adds AES and SAES for stm32c5


leftger:
~I checked the registers myself. This change is accurate. My only recommendation would be to see if you could check and see if there's any other peripherals you might work on in the future that are already in any of the YAMLs we have. My recommendation would be to see if you can bring in RNG/TRNG/PKA/CRYPTO/HASH/etc since they're interconnected within most applications. It also saves you the effort of having to do 2 PRs for each new peripheral driver you'd want to add to Embassy. No pressure on any of this, more of a suggestion since I went through this a year ago with the WBA series.~

leftger (approved):

PR #883, opened 2026-07-29
C5 crypt

usbalbin:


stm32-data-gen/src/rcc.rs
```diff
@@ -341,6 +343,7 @@ impl ParsedRccs {
             ("USB_OTG_HS", &["USB", "USBPHYC", "OTGHS", "CLK48", "ICLK"]),
             ("DTS", &["TMPSENS"]),
             ("SDMMC1", &["SDMMC1", "CLK48"]),
+            ("RNG", &["CLK48"]),
```
usbalbin, line 346: Will this break someone else?

data/registers/rng_v4.yaml
```diff
@@ -0,0 +1,262 @@
...
+enum/RNG_CONFIG2:
+  bit_size: 3
+  variants:
+  - name: Recommended
+    description: Recommended value for config A, B and C
+    value: 0
+enum/RNG_CONFIG3:
+  bit_size: 4
+  variants:
+  - name: Recommended
```
usbalbin, line 254: Any thoughts on my naming here vs `ConfigA_B_C`?

stm32-data-gen/src/rcc.rs
usbalbin: From a quick glance, this seems to produce diffs for F4, F7, G4, L0, L4, L5 and U0.

The reason I added this was because embassy would not catch me trying to run the RNG without setting up the rcc mux.

Do you know of the top of your head if this change is good or bad for all the other parts that I mess with? I'll try to check them all in time..

leftger: @xoviat might know better than me. I'll let them decide.

data/registers/rng_v4.yaml
leftger: Try running 'chiptool fmt'. If it doesn't change anything in your additions the naming should be OK

data/registers/rng_v4.yaml
```diff
@@ -0,0 +1,262 @@
...
+enum/NISTC:
+  bit_size: 1
+  variants:
+  - name: Default
+    description: Hardware default values for NIST compliant RNG
+    value: 0
+  - name: Custom
+    description: Custom values for NIST compliant RNG
+    value: 1
+enum/RNG_CONFIG1:
```
usbalbin, line 236: Not sure how it works for the other parts, but I had to add these enums manually here and fetch things from RM0522 and AN4230. So there is some extra risk of errors for this and the enums below.

leftger: If you can take a screenshot of these register values (in the RM register map) and paste them in the PR convo that should be good enough to justify the manual entries

usbalbin: image

image

image

data/registers/rng_v4.yaml
usbalbin: `chiptool fmt data/registers/rng_v4.yaml` seems happy. Assuming that is how it is used

data/registers/rng_v4.yaml
usbalbin: And HTCR:

image

stm32-data-gen/src/rcc.rs
xoviat: f732:

image

f765:

image


g441:

image

l431:

image


so broadly correct; if this breaks something we can go back and fix it later.

PR #884, opened 2026-07-30
C5: Add flash info

usbalbin:
RM0522:

image
and this also mentions 2kB
image

**Edit:** Oh, 2kB is for EDATA


stm32-data-gen/src/memory.rs
```diff
@@ -429,6 +429,7 @@ struct FlashInfo {
 #[allow(clippy::identity_op)]
 const FLASH_INFO: &[(&str, &[FlashInfo])] = &[
     ("STM32C0.*",               &[FlashInfo{ erase_value: 0xFF, write_size:  8, erase_size: &[(  2*1024, 0)] }]),
+    ("STM32C5.*",               &[FlashInfo{ erase_value: 0xFF, write_size: 16, erase_size: &[(  8*1024, 0)] }]),
```
usbalbin, line 432: Looking at https://github.com/embassy-rs/stm32-data/pull/884#issue-5019404758, is this correct?

xoviat: Not really sure about these. If it doesn't work, let me know

PR #891, opened 2026-08-11
Add TS_CAL1/TS_CAL2 temperature sensor factory calibration registers

leftger:
## Summary

Adds factory-programmed temperature sensor calibration data (TS_CAL1 at ~30°C, TS_CAL2 at ~110/130°C) as standalone peripherals, following the same pattern already used for `VREFINTCAL`, across: C0, F0, F3, F4, F7, G0, G4, H5, H7, L0, L1, L4, L4+, L5, U0, U5, WB, WBA, WL.

This picks up where #506 left off.

Addresses are sourced from ST's electronic-signature/factory-calibration reference, since most reference manuals explicitly defer the calibration address to the datasheet rather than printing it. Wherever a reference manual *does* document the address (L0x1, H7R/S, U5, WBA5/WBA6, WB0/WB05-07), values were cross-checked and one correction was made: WB0's `TS_CAL1` offset from the `DESIG` base is `+0x60`, not `+0x5C`. A real discrepancy was also confirmed between families that otherwise look similar: WL's `TS_CAL2` sits at `0x1FFF75C8`, two bytes off from WB/G0/G4/L4's `0x1FFF75CA`.

Sub-line address differences within a family (e.g. F0 value line vs mainstream, F7's two address groups, G0's mainstream line plus a value-line exception, H7's mainline vs A/B/R/S region, L1's Cat.1/2 vs Cat.3-6, WBA5 vs WBA6) are handled with non-overlapping `chip:` regex overlays so no chip ends up matching two conflicting entries for the same peripheral.

Intentionally left out (no reliable data or the register doesn't exist):
- **F1, F2** — no factory calibration register exists at all.
- **N6** — has an unrelated "Digital Temperature Sensor" (DTS) peripheral instead.
- **L0's base L010 line** — no temperature sensor at all (the rest of L0 is covered).
- **U3, WBA2** — no manual/datasheet data available yet.
- **WB0** (WB05/06/07/09xx) — the `DESIG` peripheral's absolute base address isn't confirmed in the reference manual, only inferred from a memory-map row, so left out pending a firmer source.

### DIFF
--- a/data/extra/STM32C0.yaml
+++ b/data/extra/STM32C0.yaml
@@ -1,0 +1,11 @@
+matches:
+  family: STM32C0
+peripherals:
+  # RM0490 only documents a single calibration point (TS_CAL1); there is no
+  # TS_CAL2 on this family.
+  - name: TS_CAL1
+    address: 0x1FFF7568
+    registers:
+      kind: tscal
+      version: v1
+      block: TSCAL
--- a/data/extra/STM32F0.yaml
+++ b/data/extra/STM32F0.yaml
@@ -7,3 +7,12 @@ peripherals:
       kind: vrefintcal
       version: v1
       block: VREFINTCAL
+  # TS_CAL1 exists on all F0 chips (RM0091/RM0360). TS_CAL2 only exists on the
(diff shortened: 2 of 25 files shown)


xoviat:
Please fix the merge conflict.

PR #894, opened 2026-08-20
feat(stm32-data): add DFSDM registers for all supported devices

M3gaFr3ak:
Adds DFSDM register definitions plus the RCC/trigger/perimap data needed to support the DFSDM peripheral across every STM32 family that has it.

- `dfsdm_v1.yaml`: the full set of DFSDM block variants (channel/filter counts, 3-/5-bit trigger width, ADC/DLY/HWID presence).
- `trigger.rs`: injected-trigger (JTRG) sources and DFSDM-to-TIM break connections per family.
- `perimap.rs`: regexes mapping each DFSDM chip group to its correct block.
- `header.rs`: add a DFSDM1_NS/DFSDM1_S alias so TrustZone L5 parts (L552/L562) resolve the DFSDM1 address.
- RCC: fix the H7A/B RCC layout, add the DFSDM2 kernel-clock mux, and add the H723+ (rm0468) RCC definition. Together these make DFSDM1/DFSDM2 show up as clocked peripherals (`RccPeripheral`) at all.

All DFSDM configs map to the single `dfsdm_v1` register file. ST's Cube headers name some variants `v2_1` (L5, MP15x), but the register layout is identical to v1 — "v2" only adds the optional HWID version-register block (0x7F0), modelled as a separate block and referenced only by the `_HWID` variants.

**Result:** 313 chips with DFSDM (394 peripheral instances), 0 unmapped; 13 DFSDM blocks in use, all emitted in metapac.

Refs embassy-rs/embassy#6805.

Note: the tab-tolerance fix in this PR is also submitted separately as #917, in case you'd rather land it independently first.


M3gaFr3ak:
Heads up: I pushed a few more commits after opening this for review (cleanup / addressing the DFSDM tab-parsing note). Branch should be stable now - no more pushes expected unless review feedback requires changes.

PR #911, opened 2026-09-03
STM32N6: add USB OTG RCC mapping and USBPHYC register block

sssamuelll:
## Summary

On the STM32N6, `USB1_OTG_HS` and `USB2_OTG_HS` are generated with `rcc: None`. They are the only two real peripherals on the chip without an RCC entry (the other 26 without one are DBGMCU, EXTI, RCC, UID, ICACHE, IWDG, the RISAF instances and the FDCAN RAMs, which legitimately have no clock gate). Without it, embassy-stm32 cannot implement `RccPeripheral`, and therefore `usb::Instance`, for them — so there is currently no USB on the N6.

This PR fixes that, and adds the N6 USB HS PHY controller block, in three commits:

1. **`stm32-data-gen: resolve RCC for peripherals added from extras`** — peripherals inserted by `apply_extras` never went through `match_peri_clock`, because extras are merged after `create_peripherals_for_chip` has already resolved RCC. Resolve it afterwards for whatever still has none, with the same stop-mode treatment. Idempotent for XML peripherals (same inputs, same `None`). Observed effect: the 17 `RISAF*` instances of the N6 (which come from `extra/`) now carry `AHB3ENR/RISAFEN` on HCLK3, which is what `rcc_n6.yaml` describes. No other family changes.
2. **`stm32-data: add USBPHYC register block for STM32N6`** — `usbphyc_n6.yaml`, transcribed from RM0486 Rev 4 §74.5 (pp. 3928–3934): `CR`, `TRIM1CR`, `TRIM2CR`, plus an `FSEL` enum with the three documented reference-clock values. One block serves both instances; `SELOTGDBG` is documented as USBPHYC2-only.
3. **`stm32-data: add STM32N6 USB OTG and USBPHYC RCC mapping`** — the root cause is a naming mismatch: `rcc_n6.yaml` has `OTG1EN`/`OTG1RST`, so the parser derives the key `OTG1`, while the peripheral is `USB1_OTG_HS`; no fallback covered it and the generic `^([A-Z]+)\d+$` rescue does not admit underscores. Add the aliases (`USB1_OTG_HS → OTG1, OTGPHY1`; `USB2_OTG_HS → OTG2, OTGPHY2`), and declare `OTG1PHYCTL` (0x4803FC00) and `OTG2PHYCTL` (0x480C0000) in `data/extra/STM32N6.yaml`, since the vendor XML does not list them (ST's header names them `USB1_HS_PHYC`/`USB2_HS_PHYC` at the same addresses).

Result for `STM32N657X0`:

```
USB1_OTG_HS  enable AHB5ENR/OTG1EN   reset AHB5RSTR/OTG1RST   kernel CCIPR6/OTGPHY1SEL  bus HCLK5
USB2_OTG_HS  enable AHB5ENR/OTG2EN   reset AHB5RSTR/OTG2RST   kernel CCIPR6/OTGPHY2SEL  bus HCLK5
OTG1PHYCTL   enable AHB5ENR/OTGPHY1EN  reset AHB5RSTR/OTGPHY1RST  kernel CCIPR6/OTGPHY1SEL  (usbphyc/n6)
OTG2PHYCTL   enable AHB5ENR/OTGPHY2EN  reset AHB5RSTR/OTGPHY2RST  kernel CCIPR6/OTGPHY2SEL  (usbphyc/n6)
```

## Verification

- Full regeneration diffed against a clean `main` baseline of all 1616 chips: **only the 24 `STM32N6*` files change**, plus the new `registers/usbphyc_n6.json`. The `USB_OTG_HS` entries of `STM32H7S3L8` and `STM32U5A5ZJ` are byte-identical before and after.
- `./d gen-all` succeeds; `cargo check` of the generated metapac passes for `stm32n657x0`, `stm32h7s3l8` and `stm32u5a5zj`.
- Every field of `usbphyc_n6.yaml` was checked against the RM pages (offsets and widths of all 20 fields).

## Notes for review

- The RISAF change in commit 1 is a side effect of making extras first-class for RCC. It is faithful to the RCC register (a single block-level enable bit, no reset bit), but it does mean the 17 RISAF instances share one enable bit. If that is undesirable for some reason, the alternative is an explicit exclusion, which felt worse than the generic fix — happy to change it.
- `OTGPHY*CKREFSEL` has no enum in `rcc_n6.yaml`, so the mux parser skips it, as it did before. Selecting the PHY reference clock source will need that enum if a driver wants to set it from the RCC config; out of scope here.
- The embassy-stm32 side (the `USB1_OTG_HS`/`USB2_OTG_HS` arms in `usb/otg.rs`, the `stm32n6` clock/PHY init in `usb/mod.rs`, and an example) is a separate PR that depends on this one landing in a metapac release.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015irzUk2ubW8HPRrtKyjLxb


sssamuelll:
The embassy-stm32 side that consumes this (the USB1_OTG_HS/USB2_OTG_HS arms, the PHY bring-up and an example) is in embassy-rs/embassy#6939, as a draft until this lands and the metapac tag is bumped.

PR #912, opened 2026-09-04
l4plus: add power control register 5

ant32:
stm32l4+ requires cr5 to be written to to go above 80Mhz. This commit adds a duplicate pwer_l4.yml but with cr5 added


xoviat:
just add cr5 to the existing yaml if it doesn't overlap.

PR #917, opened 2026-09-13
Fix interrupt parsing for headers using tabs in IRQn enum alignment

M3gaFr3ak:
## Symptom

DFSDM interrupts were missing from generated data for STM32F777VI, despite being correctly defined in the chip's header, correctly listed in the NVIC XML, and correctly handled by the interrupt-to-peripheral matching logic in interrupts.rs. This was the observable case that surfaced the bug, but the underlying issue is generic and can affect any interrupt on any chip.

## Root cause

stm32f777xx.h uses a tab character to align the four DFSDM1_FLTx_IRQn enum entries, while every other IRQn line in the file uses plain spaces. This looks like a formatting artifact rather than anything intentional: tabs and spaces are functionally identical to the C compiler, and the same DFSDM lines in other STM32 headers use plain spaces.

The generator's header-parsing regex in header.rs:

    r"^([a-zA-Z0-9_]+)_IRQn *= *(\d+),? "

only matches literal ASCII spaces, so it fails to match lines with a tab. The affected IRQs never enter exists_irq for that chip/core. Downstream, interrupts.rs treats any NVIC-XML IRQ not present in exists_irq as not existing on the chip and skips it. That behavior is correct in general, since NVIC XML files are shared across many chip variants, but here it fires because of a header-parsing miss rather than the peripheral being genuinely absent.

DFSDM happened to be the peripheral that exposed this, but the fix applies to the parser generically. Any header with a stray tab in this position, for any peripheral, on any chip, hits the same failure mode.

## Fix

Widen the whitespace matching in the regex to accept tabs as well as spaces (`[ \t]*` instead of ` *`), including the trailing separator before the comment. Minimal, backward-compatible change, verified against both the tab-containing and space-only header lines, with no change in behavior for already-working lines.

## Testing

- Confirmed the tab appears only on the four DFSDM lines in stm32f777xx.h.
- Manually traced the regex against both the tab and space-only formats with the fix applied. Both match correctly.
- [ ] Regenerate affected chip data and confirm DFSDM1 interrupts now appear for STM32F777 variants, with no regressions elsewhere.
- [ ] Spot-check other headers for the same tab pattern to see if other chips/peripherals were silently affected too.

## Scope and risk

Low risk. The change only broadens what already-passing lines match against, since spaces are a subset of `[ \t]`. No previously-parsed IRQ should change name or value. Recommend a full regen and diff review to catch any other headers with the same formatting issue.


Dirbaio:
nice find!

